Issue #4, opened 2019-05-27
handle protocol of provider url when init api instance


ianhe8x:
support http/https/ws/wss

PR #10, opened 2019-05-30
feat: handle protocol of provider url

AlexWZ0129:
#4

### DIFF
--- a/packages/api/src/Api.ts
+++ b/packages/api/src/Api.ts
@@ -19,12 +19,13 @@ import Alias from '@cennznet/types/alias';
 import {ApiPromise} from '@plugnet/api';
 import {ApiOptions as ApiOptionsBase} from '@plugnet/api/types';
 import {ProviderInterface} from '@plugnet/rpc-provider/types';
-import WsProvider from '@plugnet/rpc-provider/ws';
 import {isFunction, isObject} from '@plugnet/util';
+
 import * as derives from './derives';
 import getPlugins from './plugins';
 import staticMetadata from './staticMetadata';
 import {ApiOptions, IPlugin} from './types';
+import {getProvider} from './util/getProvider';
 import {injectOption, injectPlugins, mergePlugins} from './util/injectPlugin';
 import logger from './util/logging';
 
@@ -51,7 +52,7 @@ export class Api extends ApiPromise {
                 ? ({provider} as ApiOptions)
                 : ({...provider} as ApiOptions);
         if (typeof options.provider === 'string') {
-            options.provider = new WsProvider(options.provider);
+            options.provider = getProvider(options.provider);
         }
         options.metadata = Object.assign(staticMetadata, options.metadata);
         let plugins: IPlugin[] = options.plugins || [];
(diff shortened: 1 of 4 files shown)


packages/api/src/Api.ts
```diff
@@ -72,3 +73,22 @@ export class Api extends ApiPromise {
         }
     }
 }
+
+export function getProvider(providerString: string): WsProvider | HttpProvider {
+    const URL_REGEX = /(wss?|https?|WSS?|HTTPS?):\/\/([-a-zA-Z0-9@:%._\+~#=]{2,256})\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/;
```
xlc: use `/i` flag to ignore case https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/ignoreCase

Or better just use [URL](https://developer.mozilla.org/en/docs/Web/API/URL) to do the parsing

Because you already to lower case, so it won't be possible to have upper case in the match.

AlexWZ0129: Cool, thanks for the reviewing @xlc ! Right, I think it's better using the URL interface than handle this by myself. I will change it. But just not sure whether this interface is compatible with react native...

packages/api/src/Api.ts
```diff
@@ -72,3 +73,20 @@ export class Api extends ApiPromise {
         }
     }
 }
+
+export function getProvider(providerString: string): WsProvider | HttpProvider {
```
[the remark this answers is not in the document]
AlexWZ0129: Sure. Will do

packages/api/src/util/getProvider.ts
```diff
@@ -0,0 +1,37 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {assert} from '@cennznet/util';
+import {HttpProvider, WsProvider} from '@plugnet/rpc-provider';
+import {ProviderInterface} from '@plugnet/rpc-provider/types';
+
+const REGEX = /^(wss?|https?|WSS?|HTTPS?):\/\//;
```
ianhe8x: since you already convert the url string to lowercase, you can remove WSS?|HTTPS from regex

packages/api/src/util/getProvider.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function getProvider(providerString: string): ProviderInterface {
+    const providerStringLowerCase = providerString.toLowerCase();
+
+    const group = REGEX.exec(providerStringLowerCase);
+    assert(group !== null, `Missing protocol: ${providerString}`);
```
ianhe8x: `assert(group, ....` is fine

AlexWZ0129: yeah, fixed

Issue #8, opened 2019-05-27
reject the promise if connection fails during api.create()


ianhe8x:
during `Api.create()` (ws provider in this case) if connection fails, 
1) we shall reject the promise.
2) destroy the ws provider, stop it doing stuffs like reconnecting in the background.

additionally, we might add a timeout for `Api.create()`

PR #13, opened 2019-06-06
feat: reject if the connection fail, add timeout

AlexWZ0129:
#8

### DIFF
--- a/packages/api/src/Api.ts
+++ b/packages/api/src/Api.ts
@@ -28,12 +28,34 @@ import getPlugins from './plugins';
 import staticMetadata from './staticMetadata';
 import {ApiOptions, IPlugin, SubmittableExtrinsics} from './types';
 import {getProvider} from './util/getProvider';
+import {getTimeout} from './util/getTimeout';
 import {injectOption, injectPlugins, mergePlugins} from './util/injectPlugin';
 import logger from './util/logging';
 
+export const DEFAULT_TIMEOUT = 10000;
+
 export class Api extends ApiPromise {
     static async create(options: ApiOptions | ProviderInterface = {}): Promise<Api> {
-        return (new Api(options).isReady as unknown) as Promise<Api>;
+        const api = await new Api(options);
+        return withTimeout(
+            new Promise((resolve, reject) => {
+                const rejectError = err => {
+                    // Disconnect provider if API initialization fails
+                    api.disconnect();
+
+                    reject(new Error('Connection fail'));
+                };
+
+                api.isReady.then(res => {
+                    //  Remove error listener if API initialization success.
+                    (api as any)._eventemitter.removeListener('error', rejectError);
+                    resolve((res as unknown) as Api);
+                }, reject);
+
+                api.once('error', rejectError);
+            }),
+            getTimeout(options)
+        );
     }
 
     // @ts-ignore
@@ -62,6 +84,7 @@ export class Api extends ApiPromise {
             isObject(provider) && isFunction((provider as ProviderInterface).send)
                 ? ({provider} as ApiOptions)
                 : ({...provider} as ApiOptions);
+
         if (typeof options.provider === 'string') {
             options.provider = getProvider(options.provider);
         }
@@ -83,4 +106,19 @@ export class Api extends ApiPromise {
             injectPlugins(this, plugins);
         }
     }
+}
+
+function withTimeout(promise: Promise<Api>, timeoutMs: number = DEFAULT_TIMEOUT): Promise<Api> {
+    if (timeoutMs === 0) {
+        return promise;
+    }
+
+    return Promise.race<Api>([
+        promise,
+        new Promise<Api>((_, reject) => {
+            setTimeout(() => {
+                reject(new Error(`Timed out in ${timeoutMs} ms.`));
+            }, timeoutMs);
+        }),
+    ]);
 }
--- a/packages/api/src/Api.ts
+++ b/packages/api/src/Api.ts
@@ -36,7 +36,7 @@ export const DEFAULT_TIMEOUT = 10000;
 
 export class Api extends ApiPromise {
     static async create(options: ApiOptions | ProviderInterface = {}): Promise<Api> {
-        const api = await new Api(options);
+        const api = new Api(options);
         return withTimeout(
             new Promise((resolve, reject) => {
                 const rejectError = err => {
--- a/packages/api/src/Api.ts
+++ b/packages/api/src/Api.ts
@@ -108,7 +108,7 @@ export class Api extends ApiPromise {
     }
 }
 
-function withTimeout(promise: Promise<Api>, timeoutMs: number = DEFAULT_TIMEOUT): Promise<Api> {
+async function withTimeout(promise: Promise<Api>, timeoutMs: number = DEFAULT_TIMEOUT): Promise<Api> {
     if (timeoutMs === 0) {
         return promise;
     }
--- a/packages/api/src/ApiRx.ts
+++ b/packages/api/src/ApiRx.ts
@@ -53,7 +53,7 @@ export class ApiRx extends ApiRxBase {
 
         return timeoutMs === 0
             ? race(api$, rejectError)
-            : race(api$.pipe(timeout(timeoutMs === undefined ? DEFAULT_TIMEOUT : timeoutMs)), rejectError);
+            : race(api$.pipe(timeout(timeoutMs || DEFAULT_TIMEOUT)), rejectError);
     }
 
     // @ts-ignore
--- a/packages/api/src/ApiRx.ts
+++ b/packages/api/src/ApiRx.ts
@@ -24,17 +24,36 @@ import {RxResult} from '@plugnet/api/rx/types';
 import {ApiOptions as ApiOptionsBase} from '@plugnet/api/types';
 import {ProviderInterface} from '@plugnet/rpc-provider/types';
 import {isFunction, isObject} from '@plugnet/util';
-import {Observable} from 'rxjs';
+import {fromEvent, Observable, race, throwError} from 'rxjs';
+import {switchMap, timeout} from 'rxjs/operators';
 
+import {DEFAULT_TIMEOUT} from './Api';
 import * as derives from './derives';
 import staticMetadata from './staticMetadata';
 import {ApiOptions, IPlugin, SubmittableExtrinsics} from './types';
 import {getProvider} from './util/getProvider';
+import {getTimeout} from './util/getTimeout';
 import logger from './util/logging';
 
 export class ApiRx extends ApiRxBase {
     static create(options: ApiOptions | ProviderInterface = {}): Observable<ApiRx> {
-        return (new ApiRx(options).isReady as unknown) as Observable<ApiRx>;
+        const apiRx = new ApiRx(options);
+
+        const timeoutMs = getTimeout(options);
+
+        const rejectError = fromEvent((apiRx as any)._eventemitter, 'error').pipe(
+            switchMap(err => {
+                // Disconnect provider if API initialization fails
+                apiRx.disconnect();
+
+                return throwError(new Error('Connection fail'));
+            })
+        );
+        const api$ = (apiRx.isReady as unknown) as Observable<ApiRx>;
+
+        return timeoutMs === 0
+            ? race(api$, rejectError)
+            : race(api$.pipe(timeout(timeoutMs === undefined ? DEFAULT_TIMEOUT : timeoutMs)), rejectError);
     }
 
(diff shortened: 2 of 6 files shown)


packages/api/src/Api.ts
```diff
@@ -31,9 +31,26 @@ import logger from './util/logging';
...
+    static async create(provider: ApiOptions | ProviderInterface = {}): Promise<Api> {
+        const options =
+            isObject(provider) && isFunction((provider as ProviderInterface).send)
+                ? ({provider} as ApiOptions)
+                : ({...provider} as ApiOptions);
+
+        const api = await new Api(options);
+        return withTimeout(
+            new Promise((resolve, reject) => {
+                api.isReady.then(api => resolve(api)).catch(e => reject(e));
```
ianhe8x: api.isReady.then(resolve, reject);

packages/api/src/Api.ts
```diff
@@ -46,11 +63,7 @@ export class Api extends ApiPromise {
      */
     cennzxSpot?: CennzxSpot;
 
-    constructor(provider: ApiOptions | ProviderInterface = {}) {
-        const options =
-            isObject(provider) && isFunction((provider as ProviderInterface).send)
-                ? ({provider} as ApiOptions)
-                : ({...provider} as ApiOptions);
+    constructor(options: ApiOptions) {
```
ianhe8x: we still need these transition codes in constructor

packages/api/src/Api.ts
```diff
@@ -31,9 +31,26 @@ import logger from './util/logging';
 
 const Types = require('@cennznet/types');
 
+export const DEFAULT_TIMEOUT = 30000;
+
 export class Api extends ApiPromise {
-    static async create(options: ApiOptions | ProviderInterface = {}): Promise<Api> {
-        return new Api(options).isReady;
+    static async create(provider: ApiOptions | ProviderInterface = {}): Promise<Api> {
+        const options =
```
ianhe8x: you should only try to extract timeout if options is ApiOptions, don't convert it here.

packages/api/src/Api.ts
```diff
@@ -73,3 +86,19 @@ export class Api extends ApiPromise {
...
+function withTimeout(promise: Promise<any>, timeoutMs: number = DEFAULT_TIMEOUT): Promise<any> {
+    if (timeoutMs <= 0) {
+        return promise;
+    }
+
+    return Promise.race([
+        promise,
+        new Promise((reslove, reject) => {
+            const timeout = setTimeout(() => {
+                clearTimeout(timeout);
```
ianhe8x: doesn't seem necessary to clearTimeout

packages/api/src/ApiRx.ts
```diff
@@ -23,17 +23,35 @@ import {ApiOptions as ApiOptionsBase} from '@plugnet/api/types';
...
 import {getProvider} from './util/getProvider';
 import logger from './util/logging';
+
 const Types = require('@cennznet/types');
 
 export class ApiRx extends ApiRxBase {
-    static create(options: ApiOptions | ProviderInterface = {}): Observable<ApiRx> {
-        return new ApiRx(options).isReady;
+    static create(provider: ApiOptions | ProviderInterface = {}): Observable<ApiRx> {
+        const options =
```
ianhe8x: same as commented in Api

packages/api/test/e2e/api.create.e2e.ts
```diff
@@ -15,20 +15,20 @@
 import {Api} from '../../src/Api';
 import staticMetadata from '../../src/staticMetadata';
 import WsProvider from '@plugnet/rpc-provider/ws';
-import {Metadata} from '@plugnet/types';
+import {Metadata, Hash} from '@plugnet/types';
 
 describe('e2e api create', () => {
-    it('For Kauri environment - checking if static metadata is same as latest', async () => {
+    it.skip('For Kauri environment - checking if static metadata is same as latest', async () => {
```
ianhe8x: can enable these two tests

packages/api/test/e2e/api.create.e2e.ts
```diff
@@ -37,4 +37,22 @@ describe('e2e api create', () => {
...
+    });
+
+    it('should get rejected if the connection fails', async () => {
+        const incorrectEndPoint = 'wss://kauri.centrality.cloud/'
+        await expect(Api.create({provider: incorrectEndPoint})).rejects.toBeDefined();
+    });
+
+    it('should get rejected if it is not resolved in a specific period of time', async () => {
+        const endPoint = 'wss://kauri.centrality.cloud/ws?apikey=d449e2d0-868a-4f38-b977-b99e1476b7f0'
+        await expect(Api.create({provider: endPoint, timeout: 1})).rejects.toBeDefined();
```
ianhe8x: use `toThrow('specified messages')`

packages/api/src/ApiRx.ts
```diff
@@ -23,17 +23,35 @@ import {ApiOptions as ApiOptionsBase} from '@plugnet/api/types';
...
+                ? ({provider} as ApiOptions)
+                : ({...provider} as ApiOptions);
+        if (typeof options.provider === 'string') {
+            options.provider = getProvider(options.provider);
+        }
+
+        const apiRx = new ApiRx(options);
+
+        return race(
+            apiRx.isReady.pipe(timeout(options.timeout ? options.timeout : DEFAULT_TIMEOUT)),
```
ianhe8x: you don't check timeout <=0 here like in Api, why?

packages/api/src/Api.ts
xlc: resolve(api.isReady) should do the same thing

packages/api/src/types.ts
```diff
@@ -29,6 +29,7 @@ export interface ApiOptions extends Pick<ApiOptionsBase, Exclude<keyof ApiOption
      */
     provider?: ProviderInterface | string;
     plugins?: IPlugin[];
+    timeout?: number; // default 30000 ms
```
ianhe8x: comment how to specify `no timeout`

packages/api/test/e2e/api.create.e2e.ts
```diff
@@ -37,4 +37,22 @@ describe('e2e api create', () => {
...
+        expect(hash).toBeDefined();
+    });
+
+    it('should get rejected if the connection fails', async () => {
+        const incorrectEndPoint = 'wss://kauri.centrality.cloud/'
+        await expect(Api.create({provider: incorrectEndPoint})).rejects.toBeDefined();
+    });
+
+    it('should get rejected if it is not resolved in a specific period of time', async () => {
+        const endPoint = 'wss://kauri.centrality.cloud/ws?apikey=d449e2d0-868a-4f38-b977-b99e1476b7f0'
```
ianhe8x: don;'t use kauri in e2e tests

packages/api/src/ApiRx.ts
AlexWZ0129: I updated Api.ts, now timeout = 0 indicates no limit, timeout < 0 will immediately throw timeout error. This timeout originally has the same behaviour.

packages/api/test/e2e/api.create.e2e.ts
AlexWZ0129: used reject.toBeDefined because toThrow only work when the rejected value is an Error object. WebSocket connection fail returns an Event object by default. I will update Api to return an Error and using toThrow here

packages/api/src/Api.ts
AlexWZ0129: sure

packages/api/src/types.ts
AlexWZ0129: sure, added

packages/api/test/e2e/api.create.e2e.ts
AlexWZ0129: do we still want to keep this test for kauri here? I mean, Kauri is not public.

ianhe8x: ok, will remove it in another PR

packages/api/src/ApiRx.ts
```diff
@@ -23,17 +23,31 @@ import {ApiOptions as ApiOptionsBase} from '@plugnet/api/types';
...
+        const apiRx = new ApiRx(options);
+        const timeoutMs =
+            isObject(options) && isFunction((options as ProviderInterface).send)
+                ? undefined
+                : (options as ApiOptions).timeout;
+
+        return race(
+            apiRx.isReady.pipe(timeout(timeoutMs ? timeoutMs : DEFAULT_TIMEOUT)),
+            fromEvent((apiRx as any)._eventemitter, 'error').pipe(
+                switchMap(err => throwError(new Error('Connection fail')))
```
AlexWZ0129: changed to use throwError to throw an Error object here, error message `Connection fail`

packages/api/src/types.ts
```diff
@@ -29,6 +29,9 @@ export interface ApiOptions extends Pick<ApiOptionsBase, Exclude<keyof ApiOption
      */
     provider?: ProviderInterface | string;
     plugins?: IPlugin[];
+    // default timeout = 30000 ms
```
ianhe8x: default 10s

packages/api/src/Api.ts
```diff
@@ -26,14 +26,27 @@ import getPlugins from './plugins';
...
+
 export class Api extends ApiPromise {
     static async create(options: ApiOptions | ProviderInterface = {}): Promise<Api> {
-        return new Api(options).isReady;
+        const api = await new Api(options);
+        return withTimeout(
+            new Promise((resolve, reject) => {
+                api.isReady.then(resolve, reject);
+
+                api.once('error', err => {
```
ianhe8x: remove all listeners from eventemitter and disconnect provider when connection fails

packages/api/src/Api.ts
```diff
@@ -26,14 +26,35 @@ import getPlugins from './plugins';
...
+
 export class Api extends ApiPromise {
     static async create(options: ApiOptions | ProviderInterface = {}): Promise<Api> {
-        return new Api(options).isReady;
+        const api = await new Api(options);
+        return withTimeout(
+            new Promise((resolve, reject) => {
+                api.isReady.then(res => {
+                    //  Remove error listener if API initialization success.
+                    (api as any)._eventemitter.removeListener('error', rejectError);
```
ianhe8x: define rejectError before use

packages/api/src/ApiRx.ts
```diff
@@ -23,17 +23,29 @@ import {ApiOptions as ApiOptionsBase} from '@plugnet/api/types';
...
 const Types = require('@cennznet/types');
 
 export class ApiRx extends ApiRxBase {
     static create(options: ApiOptions | ProviderInterface = {}): Observable<ApiRx> {
-        return new ApiRx(options).isReady;
+        const apiRx = new ApiRx(options);
+
+        const timeoutMs = getTimeout(options);
+        return race(
+            apiRx.isReady.pipe(timeout(timeoutMs ? timeoutMs : DEFAULT_TIMEOUT)),
```
ianhe8x: 1. from what i tested, if 0 or neg number is passed to timeout, timeout will always be triggered. which is not what we want here.
2.  don't use `timeoutMs ?` to check undefined, because timeoutMs is a number, 0 will be reckon as false.

packages/api/src/ApiRx.ts
```diff
@@ -23,17 +23,29 @@ import {ApiOptions as ApiOptionsBase} from '@plugnet/api/types';
...
+            fromEvent((apiRx as any)._eventemitter, 'error').pipe(
+                switchMap(err => throwError(new Error('Connection fail')))
```
ianhe8x: use `takeUntil` to unregister it after initialization of api.

packages/api/src/Api.ts
AlexWZ0129: sure

packages/api/src/ApiRx.ts
AlexWZ0129: It's expected that timeout will be triggered if neg number is passed to timeout. And timeout should not be triggered if 0 is passed in, I tested [here](https://github.com/cennznet/api.js/blob/bf3d189272332f79829b1773eb5df7ed78d52ece/packages/api/test/e2e/rxapi.create.e2e.ts#L28)

AlexWZ0129: Ah right, will update this. I tested [here](https://github.com/cennznet/api.js/blob/bf3d189272332f79829b1773eb5df7ed78d52ece/packages/api/test/e2e/rxapi.create.e2e.ts#L28), looks like this case is passed because of using the DEFAULT_TIMEOUT when 0 is passed in...

PR #14, opened 2019-06-11
Plugin Injection

charles-salmon:
This PR ensures that:

* `genericAsset` is typed as `GenericAsset`, as opposed to `GenericAsset | undefined`.
* `cennzxSpot` is typed as `CennzxSpot`, as opposed to `CennzxSpot | undefined`.
* Consumers of this API, using TypeScript, do not need to add unnecessary guards if they have instantiated an instance of `Api` by `await`'ing a call to `Api.create()`.

### DIFF
--- a/packages/api/src/Api.ts
+++ b/packages/api/src/Api.ts
@@ -33,18 +33,43 @@ const Types = require('@cennznet/types');
 
 export class Api extends ApiPromise {
     static async create(options: ApiOptions | ProviderInterface = {}): Promise<Api> {
-        return new Api(options).isReady;
+        return new Api(options).isReady as Promise<Api>;
     }
 
     // TODO: add other crml namespaces
+    
+    private _genericAsset?: GenericAsset;
+    private _cennzxSpot?: CennzxSpot;
+
     /**
      * Generic Asset CRML extention
      */
-    genericAsset?: GenericAsset;
+    get genericAsset(): GenericAsset {
+        if (!this._genericAsset) {
+            throw new Error('Generic Asset plugin has not been injected.');
+        }
+
+        return this._genericAsset;
+    }
+
+    set genericAsset(value: GenericAsset) {
+        this._genericAsset = value;
+    }
+
     /**
      * Cennzx Spot CRML extention
      */
-    cennzxSpot?: CennzxSpot;
+    get cennzxSpot(): CennzxSpot {
+        if (!this._cennzxSpot) {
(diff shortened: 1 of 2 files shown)


packages/api/src/Api.ts
```diff
@@ -33,18 +33,43 @@ const Types = require('@cennznet/types');
...
     // TODO: add other crml namespaces
+    
+    private _genericAsset?: GenericAsset;
+    private _cennzxSpot?: CennzxSpot;
+
     /**
      * Generic Asset CRML extention
      */
-    genericAsset?: GenericAsset;
+    get genericAsset(): GenericAsset {
```
ianhe8x: we use `Object.defineProperty` to set genericAsset and cennzxSpot, so i think you can just throw Error in the getter, (without creating and visiting `_genericAsset`)

packages/api/src/Api.ts
```diff
@@ -33,18 +33,43 @@ const Types = require('@cennznet/types');
 
 export class Api extends ApiPromise {
     static async create(options: ApiOptions | ProviderInterface = {}): Promise<Api> {
-        return new Api(options).isReady;
+        return new Api(options).isReady as Promise<Api>;
```
ianhe8x: the return type already have `Promise<Api>`, my ide can recognize it correctly. type casting doesn't seem necessary

charles-salmon: `isReady` returns a `Promise<ApiPromise>` (and `ApiPromise` does not have a `cennzxSpot` or `genericAsset` property), so I believe the cast is necessary.

ianhe8x: image
create() has a return type defined.

packages/api/src/Api.ts
charles-salmon: That sounds like a fair optimisation to me!

packages/api/src/Api.ts
charles-salmon: Yes, it does. But that still doesn't negate the need for this cast.

See:

Screenshot from 2019-06-12 10-52-45

ianhe8x: i don't use vs-code, but sure we can add this if it helps.

charles-salmon: Just to be clear, this isn't VSCode specific.

The `isReady` method returns a `Promise<ApiPromise>`. The `create()` method specifies a return type of `Promise<Api>`. `Api` contains properties that `ApiPromise` does not, which is why the case is **required**.

PR #36, opened 2019-08-22
chore: adding dependency

AlexZhenWang:


ianhe8x:
this dependency is not needed after upgrade to using `@plugnet/api@0.90.x`

PR #49, opened 2019-09-18
remove provider instance line from readme

KarishmaBothara:


ianhe8x:
That is still allowed, when user don't want the default implementation. for example user can use a AdvancedProvider which supports http/socks proxy.

PR #52, opened 2019-10-22
fix: change peerDependencies to dependencies in crml-attestation/package.json

ianhe8x:
A missing piece should be done when I put crml-attestation into this repo.


benxgao:
@ianhe8x Could you please add @KarishmaBothara as a reviewer as well?

PR #53, opened 2019-10-23
Fee changes

KarishmaBothara:


packages/api/src/Api.ts
```diff
@@ -116,6 +116,7 @@ export class Api extends ApiPromise {
 
     decorateCennznetExtrinsics(): void {
         decorateExtrinsics(this);
+        decorateExtrinsicsSubmittables(this);
```
ianhe8x: can you combine them together? decorateExtrinsics and decorateExtrinsicsSubmittables

packages/api/src/util/filterEvents.ts
```diff
@@ -0,0 +1,35 @@
...
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import {EventRecord, SignedBlock} from '@plugnet/types/interfaces';
+
+import {U8a} from '@plugnet/types';
+
+import l from './logging';
+
+export default function filterEvents(
```
ianhe8x: I would like the utils you added moved to `submittable/utils/*`, so we can easily remove them when we don't need this hack someday in the future

packages/api/src/Api.ts
KarishmaBothara: yes

packages/api/src/util/filterEvents.ts
KarishmaBothara: Makes sense..done

packages/api/src/types.ts
```diff
@@ -31,12 +32,25 @@ import ApiBase from '@plugnet/api/base';
 import {
     ApiOptions as ApiOptionsBase,
     SignerOptions as SignerOptionsBase,
+    // SubmitableResultResult,
```
ianhe8x: clean these comments

packages/api/src/util/customDecorators.ts
```diff
@@ -23,3 +32,12 @@ function decorateExtrinsic(
     };
     return (api as any).decorateFunctionMeta(method, newMethod);
 }
+
+function decorateExtrinsicsSubmittable(
```
ianhe8x: this method can be combined with decorateExtrinsic.
replace `const extrinsic = method(...params);` with `const extrinsic = creator(method(...params));`

packages/api/src/util/customDecorators.ts
```diff
@@ -1,21 +1,26 @@
...
+            api.tx[sectionName][methodName] = decorateExtrinsic(api, api.tx[sectionName][methodName], creator);
         }
     }
 }
 
 function decorateExtrinsic(
     api: Api | ApiRx,
-    method: SubmittableExtrinsicFunction<any>
+    method: SubmittableExtrinsicFunction<any>,
+    creator: any
```
ianhe8x: can we use a better type than `any`

packages/api/src/types.ts
```diff
@@ -28,15 +29,14 @@ import {
...
 import {
-    ApiOptions as ApiOptionsBase,
-    SignerOptions as SignerOptionsBase,
     SubmitableResultResult,
     SubmitableResultSubscription,
     SubmittableExtrinsic as SubmittableExtrinsicBase,
     SubmittableResultImpl,
-    UnsubscribePromise,
-} from '@plugnet/api/types';
+} from '@cennznet/api/submittable/types';
```
ianhe8x: should use relative path here, since what you import here is within same package.

packages/api/src/util/customDecorators.ts
```diff
@@ -1,21 +1,26 @@
 import {Api, ApiRx} from '@cennznet/api';
+import createSubmittable from '@cennznet/api/submittable/createSubmittable';
```
ianhe8x: same problem, should use relative path. and some old codes have this issue as well, pls correct them together.

packages/api/src/util/customDecorators.ts
```diff
@@ -1,21 +1,26 @@
...
+import {SubmittableExtrinsic} from '@cennznet/api/submittable/types';
 import {SubmittableExtrinsicFunction} from '@cennznet/api/types';
 import {CodecArg} from '@cennznet/types/types';
+import {CodecArg as Arg} from '@plugnet/types/types';
```
ianhe8x: I think this Arg is the same as CodecArg from '@cennznet/types/types'

packages/api/src/types.ts
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {DecoratedCennznetDerive} from '@cennznet/api/derives';
+//import ApiBase from '@cennznet/api/base1';
```
ianhe8x: unwanted comments

packages/api/src/util/customDecorators.ts
```diff
@@ -1,21 +1,26 @@
-import {Api, ApiRx} from '@cennznet/api';
-import {SubmittableExtrinsicFunction} from '@cennznet/api/types';
 import {CodecArg} from '@cennznet/types/types';
+import {Call} from '@plugnet/types/interfaces';
+import {Api, ApiRx} from '../../src';
```
ianhe8x: this path goes too deep

packages/api/src/submittable/createSubmittable.ts
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable no-dupe-class-members */
+// Copyright 2017-2019 @polkadot/api authors & contributors
```
benxgao: Is this file from `polkadot` or owned by centrality?

packages/api/src/submittable/Submittable.ts
```diff
@@ -0,0 +1,293 @@
+/* eslint-disable no-dupe-class-members */
+// Copyright 2017-2019 @polkadot/api authors & contributors
```
benxgao: Correct copyright?

packages/api/src/submittable/createSubmittable.ts
KarishmaBothara: This is taken from plugnet. May be I need to add our copyright,

benxgao: According to the conversation in [this thread](https://centralityteam.slack.com/archives/CNFU9UTM3/p1571338509043200), it seems it has not come across to any decision yet.

ianhe8x: we didn't make any changes on this file, so we should keep the header untouched

packages/api/src/submittable/Submittable.ts
```diff
@@ -0,0 +1,307 @@
...
+    SubmittableExtrinsic,
+    SubmittableResultImpl,
+} from './types';
+
+import {ClassOf, createType, Vec} from '@plugnet/types';
+import {isBn, isFunction, isNumber, isUndefined} from '@plugnet/util';
+import {combineLatest, Observable, of} from 'rxjs';
+import {first, map, mergeMap, switchMap, tap} from 'rxjs/operators';
+
+//import ApiBase from '../base1';
```
ianhe8x: remove unnecessary comments

packages/api/src/util/customDecorators.ts
```diff
@@ -1,21 +1,40 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 import {CodecArg} from '@cennznet/types/types';
+import {Call} from '@plugnet/types/interfaces';
+import {Api, ApiRx} from '..';
```
ianhe8x: might create circular dependency, it would be better if you import them from their own file.

PR #58, opened 2019-10-30
v0.20.8

KarishmaBothara:


packages/api/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.20.7]() (2019-10-30)
```
AlexZhenWang: I guess this should be `# [0.20.7]` not `## [0.20.7]`? I mean, the one below is like `# [0.20.6]`

KarishmaBothara: I made changes for 0.20.8 so I added it for 0.20.8.. 
I think 0.20.7 is missing.

AlexZhenWang:
LGTM

benxgao:
As we have implemented new branching model, we gonna make tags and releases following new approach

PR #69, opened 2019-12-02
Update dependency

KarishmaBothara:
Changed all plugnet dependency with polkadot stable non beta version and made the build work. Able to connect to node(local) using cennznet node’s develop branch and query storage.


packages/types/src/injects.ts
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//import {Balance} from '@polkadot/types/interfaces';
```
holygits: Can remove this

packages/api/test/e2e/api.create.e2e.ts
```diff
@@ -32,8 +32,8 @@ describe('e2e api create', () => {
     });
 
     it('should create an Api instance with the timeout option', async () => {
-        const endPoint = 'wss://rimu.unfrastructure.io/public/ws';
-        api = await Api.create({provider: endPoint, timeout: 1000000000});
+        // const endPoint = 'wss://rimu.unfrastructure.io/public/ws';
```
holygits: Should move this `endPoint` into a constant / config file and re-use appropriately.
Also make a note that we're testing against the local cennznet develop branch
I'll add a task to re-instate to re-configure the endpoint when a live network is up

packages/api/test/e2e/fees.e2e.ts
```diff
@@ -59,15 +47,29 @@ describe('fees in cennznet', () => {
             });
             const fee = await tx.fee(sender.address);
```
holygits: Maybe this can be renamed to `calculatedFee` (I know you didn't write this originally but can fix since we're changing this already).

`gas` isn't really correct name either should be `actualFee` or something.

So then the test is `expect(calculatedFee.toString()).toEqual(actualFee.toString());`

Same with the other usages

packages/api/test/e2e/fees.e2e.ts
```diff
@@ -95,8 +96,18 @@ describe('fees in cennznet (Rxjs)', () => {
                 initialIssuance: 100,
             });
             const fee = await tx.fee(sender.address).toPromise();
+            console.log('fee', fee.toString());
```
holygits: Can remove this

packages/api/test/e2e/fees.e2e.ts
```diff
@@ -50,12 +50,8 @@ describe('fees in cennznet', () => {
...
                 const blockHash = status.asFinalized;
                 const events = ((await api.query.system.events.at(blockHash)) as unknown) as Vec<EventRecord>;
-                let gas;
-                for (const {event} of events) {
-                  if (event.method === 'Charged') {
-                    gas = event.data[1];
-                  }
-                }
+                const feeChargeEvent = events.find(event => event.event.data.method === 'Charged');
+                const gas = feeChargeEvent.event.data[1];
```
holygits: `gas` is not the right term, can change to `realFee` or similar (I know you didn't write this originally but since we're changing it already)

The expect should read more like:
` expect(realFee.toString()).toEqual(calculatedFee.toString());`
same goes for the other usages

packages/api/test/e2e/tx.e2e.ts
```diff
@@ -36,7 +37,9 @@ describe('e2e transactions', () => {
     let api: Api;
 
     beforeAll(async () => {
-        api = await Api.create({provider: 'wss://rimu.unfrastructure.io/public/ws'});
+        // api = await Api.create({provider: 'wss://rimu.unfrastructure.io/public/ws'});
```
holygits: Can remove this now

packages/api/constants.ts
```diff
@@ -0,0 +1,22 @@
+import {ApiOptions} from '@cennznet/api/types';
+
+export type NetworkConfig = {
+  options: ApiOptions,
+  defaultEndpoint: string
+};
+
+export type NetworkName = 'CUSTOM' | 'RIMU' | 'KAURI';
```
benxgao: I would suggest to put `configuring network options in a separated ticket / PR`, as this is not quite related to `update dependency`.

benxgao: As per `configuring network`, I have a different solution, which sits in the [demo of automation-test](https://github.com/benxgao/substrate-automation-test/tree/master/config), if you are available, would you mind taking a look, or we may have a discussion if needed?

PR #70, opened 2019-12-06
Update extrinsic v2

KarishmaBothara:

PR #71, opened 2019-12-11
[WIP] Set up private testnet with circleci integration test

benxgao:


scripts/circleci/run_automation_test.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# https://stackoverflow.com/questions/16483119/an-example-of-how-to-use-getopts-in-bash
+
+set -ex
+
+# usage sample:
+# ./scripts/circleci/run_automation_test.sh -t integration
```
benxgao: Here is how to start with integration test locally

scripts/clear_testnet.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
benxgao: To clear testnet locally

Dockerfile.integration
```diff
@@ -0,0 +1,31 @@
+FROM node:13
```
holygits: Dockerfile is the extension so it should be named `integration.Dockerfile`

jest/unit.config.js
```diff
@@ -0,0 +1,34 @@
...
+    '^.+\\.(js|jsx)$': 'babel-jest',
+    '^.+\\.(ts|tsx)$': 'ts-jest',
+  },
+  testMatch: ['/**/?(*.)+(spec).[jt]s?(x)'],
+  // testEnvironment: 'node',
+  // testEnvironment: './jest/env.js',
+  moduleNameMapper: {
+    '@cennznet/api(.*)$': '<rootDir>/packages/api/src/$1',
+    '@cennznet/crml-attestation(.*)$': '<rootDir>/packages/crml-attestation/src/$1',
+    '@cennznet/crml-generic-asset(.*)$': '<rootDir>/packages/crml-generic-asset/src/$1',
```
holygits: Maybe ignore `crml-*` runtime packages from this initially since it's not required for #70

config/development.js
```diff
@@ -0,0 +1,7 @@
+import R from 'ramda';
+
+import defaultConfig from './default';
+
+const config = {};
+
+export default R.mergeDeepRight(defaultConfig, config);
```
holygits: This is neat. Could add a comment explaining what the expected `config` should look like after `mergeDeepRight` occurs, since it's not immediately obvious or a unit test?

docker-compose.yml
```diff
@@ -1,93 +1,78 @@
 version: '3'
 services:
-  node-0:
-    image: cennznet/cennznet:0.9.24-beta-639f5f36
+  node_alice:
```
holygits: This should be simplified to a single node.
The reasoning being the functionality of the API under test should not change with the number of nodes on the network.
- `volumes` and  mounting stuff is not required here since nothing should be persisted between CI test runs
- I don't think a UI server needs to be spun up as part of this CI pipeline

This file should be really simple with two services a CENNZnet node container and an API container to run the e2e tests
```yaml
services:
  cennznet-node:
    image: cennznet/cennznet:rc1
    command:
      - -- dev
    ports:
      - "9933:9933"
      - "9944:9944"
  cennznet-api:
   # ...
```

jest/integration.config.js
```diff
@@ -0,0 +1,34 @@
...
+  transform: {
+    '^.+\\.(js|jsx)$': 'babel-jest',
+    '^.+\\.(ts|tsx)$': 'ts-jest',
+  },
+  testMatch: ['/**/?(*.)+(test).[jt]s?(x)'],
+  // testEnvironment: 'node',
+  // testEnvironment: './jest/env.js',
+  moduleNameMapper: {
+    '@cennznet/api(.*)$': '<rootDir>/packages/api/src/$1',
+    '@cennznet/crml-attestation(.*)$': '<rootDir>/packages/crml-attestation/src/$1',
```
holygits: `crml-*` packages could be removed from the initial jest config since they aren't required for #70 or maybe the actual packages should be removed in #70

Dockerfile.integration
holygits: Dockerfile is the extension so it should be named `integration.Dockerfile`

config/development.js
benxgao: Right! Also `config` here is only used in testing, so `ramda` should be fine. but later if we want to adopt it into api, may need to replace ramda with a smaller package in considering of sdk bundling size~

docker-compose.yml
benxgao: Yes, The UI together with multiple nodes was to demo isolated test network in docker, but definitely not need in CI,    will simplify it

jest/integration.config.js
benxgao: `moduleNameMapper ` are copied from the current legacy settings.

Could comment them for now, and remove `crml-*` together with testing stuff in a separated PR, as current PR is already so big

jest/unit.config.js
benxgao: Similar to the above concern~

Dockerfile.integration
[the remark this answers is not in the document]
benxgao: Agreed.

PR #89, opened 2020-03-05
Support CENNZnet node's develop branch

KarishmaBothara:
Checked all test

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3'
 services:
   testnet_node_alice:
     container_name: testnet_node_alice
-    image: cennznet/cennznet:latest
+    image: cennznet/cennznet@sha256:8a84ab9366c3be3c5d2062b5df52969179840addfd58f289608bfc612754b6c3
     command:
       - --chain=dev
       - --base-path=/mnt/node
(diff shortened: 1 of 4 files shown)


packages/types/src/injects.ts
```diff
@@ -19,6 +19,7 @@ export default {
   ...runtimeTypes,
   ...extrinsicTypes,
   AssetOf: 'u128',
+  Address: 'AccountId',
```
benxgao: Does it work with current rc2 if we add this type?

docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 services:
   testnet_node_alice:
     container_name: testnet_node_alice
-    image: cennznet/cennznet:1.0.0-rc2
+    image: cennznet/cennznet:latest
```
benxgao: Would it be better to stay with rc2, only if cennznet:latest is always stable, otherwise, api automation tests may be failed? @holygits

holygits: we really should tag with a commit as well in cennznet but for now workaround is either use sha hash of image so it won't update: `cennznet/cennznet@sha256:8a84ab9366c3be3c5d2062b5df52969179840addfd58f289608bfc612754b6c3` or I don' think circleci will re-pull the latest image anyway

packages/types/src/injects.ts
KarishmaBothara: no

docker-compose.yml
KarishmaBothara: added Sha hash of image

PR #91, opened 2020-03-09
Remove deprecated method from crml-cennzx-spot repo

KarishmaBothara:
Removed inputPrice and outputPrice methods.


benxgao:
I am not sure if there were any new functions added in new [cennzx-spot](https://github.com/cennznet/cennznet/blob/110f4bd24466385ef7dbb8f3bbd99b9215cd6f94/crml/cennzx-spot/src/lib.rs), but if there was anything we need wring more tests in api side, we could make a separate PR on it, otherwise, LGTM.

packages/crml-cennzx-spot/src/CennzxSpot.ts
```diff
@@ -18,229 +18,202 @@ import {AssetId} from '@cennznet/types';
...
+
+    return _fn;
+  }
+
+  /**
+   * Query liquidity price for a core asset amount
+   * @param assetId The id of the asset
+   * @param coreAmount - the amount of core asset
+   */
+  get liquidityPrice(): QueryableGetLiquidityBalancePrice {
```
holygits: Random note: It's weird that some methods are prefixed **get**`TheThingToQuery` while some are just `thingToQuery`
We could improve this

benxgao:
> Changes look good so far.
> 
> Some inconsisteny in the API is noticeable from these clean ups e.g.
> `api.rpc.cennzx.buyPrice` vs. `api.cennzxSpot.getPoolCoreAssetBalance` is confusing for a couple reasons
> why is one under `api.rpc.cennzx` and the other under `api.cennzxSpot` and why is one called getPool... while the other is called buyPrice when both are 'gets'
> 
> Let's address this in a follow up PR
> I'm aware these changes aren't because of this PR so am keen to accept this but let's follow up with a clean up PR

I assume any storage/methods defined in `cennzxSpot` module in node could be retrieved via `api.cennzxSpot.xx` on api side? And except rpc calls, which are not directly available and we have to decorate on api side? Want to understand more in this area, thanks @holygits

holygits:
We create the alias `api.cennzxSpot` and all the function names with 'getWhatEver` in the derive package. The metadata generated names are `api.query.cennzxSpot.totalBalance` etc. so it's something we can tidy up here

PR #93, opened 2020-03-16
Added derived query and test case to get estimated fee

KarishmaBothara:
This PR enables users to know an extrinsic’s fee in a particular currency they wish to pay in.
Files changed -

Added ~
fees (folder inside derives)
estimateFee.ts - (file which has the derived query)
index.ts - (All derived queries will be included from this file)

Modified ~
cennzx.e2e.ts (added new test cases to cover fee exchange [positive/negative flow])
queries.e2e.ts (updated the earlier test case with correct estimate parameter [partialFee over weight])
constants.ts (added DEFAULT_FEE_ASSET id)

A very interesting observation while working on this PR was when we use signFake it takes all the default parameters (signatureOption) so if we check packages/types/src/extrinsic/v2/ExtrinsicSignature.ts

createPayload(
    method: Call,
    {
      blockHash,
      era,
      genesisHash,
      nonce,
      doughnut,
      runtimeVersion: {specVersion},
      tip,
      transactionPayment,
    }: ExtrinsicV2SignatureOptions
  ): ExtrinsicPayloadV2 {
    return new ExtrinsicPayloadV2(this.registry, {
      blockHash,
      doughnut: doughnut || createType(this.registry, 'Option<Doughnut>'),
      **_era: era || IMMORTAL_ERA,_**
      genesisHash,
      method: method.toHex(),
      nonce,
      specVersion,
      // [[tip]] is now set inside [[transactionPayment]]
      // This doesn't do anything, just signalling our intention not to use it.
      tip: null,
      transactionPayment: transactionPayment || createType(this.registry, 'ChargeTransactionPayment'),
    });
  }

Here era if not supplied will take IMMORTAL ERA which is of encoded length 1.
And if we check the document on polkadot site, it says
https://polkadot.js.org/api/start/api.tx.html#under-the-hood
The API will retrieve the current block hash and use it to create a mortal transaction, i.e. the transaction will only be valid for a limited number of blocks (by default this is 5 mins at 6s block times)

So default is always Mortal ERA..  Even on polkadot.js api they use IMMORTAL_ERA as default.
This leads to change the encoded length of payload and incorrect fees.

In this PR, just to fix the encoded length of payload, we create ExtrinsicEra as Mortal era.


packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,61 @@
...
+          });
+          return api.rpc.payment.queryInfo(extrinsic.toHex());
+        }
+      ),
+      switchMap(paymentInfo => {
+        const feeInBaseCurrency = paymentInfo.partialFee;
+        if (feeAssetId.toString() === DEFAULT_FEE_ASSET) {
+          return of(feeInBaseCurrency);
+        } else {
+          return (api.rpc as any).cennzx.buyPrice(DEFAULT_FEE_ASSET, feeInBaseCurrency, feeAssetId);
```
holygits: Maybe it's more clear to call these user and network e.g. `feeAssetId` -> `userFeeAssetId` and `DEFAULT_FEE_ASSET` -> `networkFeeAssetId`

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,61 @@
...
+import ExtrinsicEra from '@polkadot/types/primitive/Extrinsic/ExtrinsicEra';
+import BN from 'bn.js';
+import {combineLatest, Observable, of} from 'rxjs';
+import {catchError, first, map, switchMap} from 'rxjs/operators';
+
+export function estimateFee(api: ApiInterfaceRx) {
+  return (
+    extrinsic: IExtrinsic,
+    feeAssetId: AnyAssetId,
+    sender: Address,
```
holygits: Also would be good to leave a comment about why we're taking these parameters i.e. to estimate the fee correctly using extrinsic signature

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,61 @@
...
+      api.rpc.chain.getBlockHash(),
+      api.rpc.chain.getBlockHash(0),
+      api.query.system.accountNonce(sender),
+    ]).pipe(
+      first(),
+      switchMap(
+        ([runtimeVersion, blockHash, genesisHash, nonce]): Observable<RuntimeDispatchInfo> => {
+          const registry = new TypeRegistry();
+          const era = new ExtrinsicEra(registry, {
+            current: 1,
```
holygits: Is this still correct at say block 10,000?

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,61 @@
...
+import BN from 'bn.js';
+import {combineLatest, Observable, of} from 'rxjs';
+import {catchError, first, map, switchMap} from 'rxjs/operators';
+
+export function estimateFee(api: ApiInterfaceRx) {
+  return (
+    extrinsic: IExtrinsic,
+    feeAssetId: AnyAssetId,
+    sender: Address,
+    transactionPayment: ChargeTransactionPayment
```
holygits: Did you try faking `transactionPayment` with some high number?

packages/api/src/derives/fees/estimateFee.ts
KarishmaBothara: changed

packages/api/src/derives/fees/estimateFee.ts
KarishmaBothara: yes, tested setting current to 10,000

packages/api/src/derives/fees/estimateFee.ts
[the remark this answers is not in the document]
KarishmaBothara: have faked sender, since we need it as parameter in signFake

KarishmaBothara: added comment

packages/api/src/derives/fees/estimateFee.ts
KarishmaBothara: yes.. it works well provided the maxPayment is same while estimating and while send an extrinsic.

KarishmaBothara:
> The api for estimate fee should really be `estimateFee(extrinsic, asset_id)`
> we can preferably fake or query as second option to get any additional info

Made modification, used fake values for others to determine the 'estimated fee'

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,66 @@
...
+import {Address} from '@polkadot/types/interfaces';
+import {RuntimeDispatchInfo} from '@polkadot/types/interfaces/rpc';
+import ExtrinsicEra from '@polkadot/types/primitive/Extrinsic/ExtrinsicEra';
+import BN from 'bn.js';
+import {combineLatest, Observable, of} from 'rxjs';
+import {catchError, first, map, switchMap} from 'rxjs/operators';
+
+export function estimateFee(api: ApiInterfaceRx) {
+  /* To estimate fee correctly using extrinsic's fake signature */
+  return (extrinsic: IExtrinsic, userFeeAssetId: AnyAssetId, maxPayment?: string): Observable<any> => {
```
KarishmaBothara: have taken maxPayment as optional parameter, cause if the maxPayment value is different while fake signing and actually signing, then the fee estimated is not the same as deducted.

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,66 @@
...
+          const sender = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
+          const feeExchange = {
+            assetId: userFeeAssetId,
+            maxPayment: maxPayment,
+          };
+          const transactionPayment =
+            userFeeAssetId.toString() === networkFeeAssetId.toString()
+              ? null
+              : {
+                  tip: 0,
```
holygits: We should make a utility constructor for FeeExchange

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,66 @@
...
+import {TypeRegistry} from '@polkadot/types';
+import {Address} from '@polkadot/types/interfaces';
+import {RuntimeDispatchInfo} from '@polkadot/types/interfaces/rpc';
+import ExtrinsicEra from '@polkadot/types/primitive/Extrinsic/ExtrinsicEra';
+import BN from 'bn.js';
+import {combineLatest, Observable, of} from 'rxjs';
+import {catchError, first, map, switchMap} from 'rxjs/operators';
+
+export function estimateFee(api: ApiInterfaceRx) {
+  /* To estimate fee correctly using extrinsic's fake signature */
```
holygits: Something more descriptive would be good e.g.
// We generate fake signature data here to ensure the estimated fee will correctly match the fee paid when the extrinsic is signed by a user.
// This is because fees are currently based on the byte length of the extrinsic

KarishmaBothara: added exactly above comments

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,56 @@
...
+      first(),
+      switchMap(
+        ([runtimeVersion, blockHash, genesisHash, networkFeeAssetId]): Observable<[RuntimeDispatchInfo, any]> => {
+          const registry = new TypeRegistry();
+          const era = new ExtrinsicEra(registry, {
+            current: 1,
+            period: 1,
+          });
+          const nonce = null;
+          const sender = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
```
benxgao: Could assign the address to `const ALICE = xxx; sender = ALICE;`, or alternatively add a comment at the end to reflect account name?

packages/types/src/runtime/transaction-payment/TransactionPayment.ts
```diff
@@ -0,0 +1,15 @@
+import {AnyAssetId} from '@cennznet/types/types';
+
+export function generateTransactionPayment(tip: number, assetId: AnyAssetId, maxPayment: string) {
```
benxgao: We do not actually need this function?  It increases code complexity, and we need to take extra effort on maintaining the relationship between `transactionPayment` and `types` packages. 

When we use it, its not quite clearly on what's meaning of each arguments like `generateTransactionPayment(0, 1, 2);`
Whats more, `types` package is more about typing, and this code segment is kinda data restructuring.

Would suggest to remove this function, and use transactionPayment original data structure directly, like:
```js
const transactionPayment = {
  // tip: tip, // we do not need to assign a value if its value is 0 by default?
  feeExchange: {
    FeeExchangeV1: {
      assetId,
      maxPayment,
    },
  },
};
```

packages/api/test/e2e/cennzx.e2e.ts
```diff
@@ -87,6 +129,18 @@ describe('CENNZX e2e queries/transactions', () => {
         );
         done();
       });
+
+      describe('feeExchange derive queries with negative flow', () => {
```
benxgao: Not a big deal, but this test is not part of `Negative flow with no liquidity in pool`, could get it in the same level of `describe('Negative flow with no liquidity in pool')`?

packages/api/src/derives/fees/estimateFee.ts
KarishmaBothara: Here it is not important to have ALICE's address or BOB's address.. doing this will create confusion as in, it is important to have ALICE's address as sender address.. Here it can be any address.

packages/types/src/runtime/transaction-payment/TransactionPayment.ts
KarishmaBothara: @holygits ? 
I don't mind either way... If we can all come to one thing, I can implement that..

packages/api/test/e2e/cennzx.e2e.ts
KarishmaBothara: It is part of negative flow with no liquidity in pool.. This derive query will result in error, only when there is no enough liquidity in the pool

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,56 @@
...
+import {AnyAssetId, IExtrinsic} from '@cennznet/types/types';
+import {drr} from '@polkadot/rpc-core/rxjs';
+import {TypeRegistry} from '@polkadot/types';
+import {RuntimeDispatchInfo} from '@polkadot/types/interfaces/rpc';
+import ExtrinsicEra from '@polkadot/types/primitive/Extrinsic/ExtrinsicEra';
+import BN from 'bn.js';
+import {combineLatest, Observable, of} from 'rxjs';
+import {catchError, first, map, switchMap} from 'rxjs/operators';
+
+export function estimateFee(api: ApiInterfaceRx) {
```
benxgao: Would suggest add some comment to describe what this interface is about, and how and when it can be used, etc. So that our API consumers do not need to look through the code and tests to grab the whole idea. 

If possible, it is even better to describe what is the currency of `estimateFee`, with giving some examples. And we eventually need to get it documented somewhere.

KarishmaBothara: Have added these lines,
// We generate fake signature data here to ensure the estimated fee will correctly match the fee paid when the extrinsic is signed by a user.
   // This is because fees are currently based on the byte length of the extrinsic

KarishmaBothara: May be I can add, this interface is do determine fee estimate for any transaction that is going to be executed.. The estimate can be in any currency provided there is enough liquidity in the exchange pool for that currency.
Do you think this is enough?

packages/api/src/derives/fees/estimateFee.ts
benxgao: Not a big deal and no need to change here, but I am thinking what if we need to add more options as arguments passed into the function?

Would be in favor of scoping args in a key-valued object whenever we need to add more than 2 args in a function, otherwise its quite hard for dev to know what is the meaning of each argument, for this one, its like:

```js
({
  extrinsic, 
  userFeeAssetId, 
  maxPayment
}) => {...}
```
```js
// Or
(extrinsic, userFeeAssetId, {
  maxPayment?: string,
  // more options comes here
}) => {... }
```
So that when this function is called, we as developer can easily identify what is what. Thats just my own opinion, there may be some other cases I did not get, welcome a discuss~

packages/api/src/derives/fees/estimateFee.ts
holygits: maybe call it `fake_sender` to avoid confusion

packages/types/src/runtime/transaction-payment/TransactionPayment.ts
holygits: It's not good to have the construction logic every place we need this type, maybe doing the KV thing helps the readability of the args e.g.: `export function makeTransactionPayment({tip: number, assetId: AnyAssetId, maxPayment: string})`

KarishmaBothara: @benxgao - are you okay with a function with argument passed as key-value pair in types package?

packages/api/test/e2e/cennzx.e2e.ts
```diff
@@ -62,6 +64,46 @@ describe('CENNZX e2e queries/transactions', () => {
...
+              const assetBalanceAfter: Balance = await api.query.genericAsset.freeBalance(CENTRAPAY, alice.address);
+              expect(assetBalanceBefore.sub(assetBalanceAfter).toString()).toEqual(feeFromQuery.toString());
+              done();
+            }
+          });
+        });
+
+        it('Query estimated fee in different currency (CENNZ)', async done => {
+          const maxPayment = '50000000000000000';
+          const transactionPayment = generateTransactionPayment({tip:0, assetId:CENNZ, maxPayment});
```
holygits: 👍

packages/api/src/derives/fees/estimateFee.ts
benxgao: A simple code example or a scenario case would explain more clearly, so that developers do not need to check test cases to see how to use it. Maybe we could add that in the future, no more struggling here if you think it is unnecessary :)

packages/api/test/e2e/cennzx.e2e.ts
benxgao: Gotcha 👍

packages/types/src/runtime/transaction-payment/TransactionPayment.ts
```diff
@@ -0,0 +1,15 @@
+import {AnyAssetId} from '@cennznet/types/types';
```
benxgao: it is still in `types` package, but it is a kind a util function (and cennznet/utils seems not needed)... 
What if just place it in the file of `packages/api/src/derives/fees/estimateFee.ts`?

Otherwise, potentially new onboard developers would add more similar utils functions into `types` package? Then the package gonna be out of control and we potentially need to make another round refactoring... :)

KarishmaBothara: @holygits ?

benxgao: Except this point, it looks good to me. 

Any inputs on what kind of functions we should consider to put inside `types` package? @holygits

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -0,0 +1,72 @@
...
+import {AnyAssetId, IExtrinsic} from '@cennznet/types/types';
+import {drr} from '@polkadot/rpc-core/rxjs';
+import {TypeRegistry} from '@polkadot/types';
+import {RuntimeDispatchInfo} from '@polkadot/types/interfaces/rpc';
+import ExtrinsicEra from '@polkadot/types/primitive/Extrinsic/ExtrinsicEra';
+import BN from 'bn.js';
+import {combineLatest, Observable, of} from 'rxjs';
+import {catchError, first, map, switchMap} from 'rxjs/operators';
+
+// This interface is to determine fee estimate for any transaction that is going to be executed..
```
benxgao: 👍

packages/types/src/runtime/transaction-payment/TransactionPayment.ts
holygits: The constructor should go with the type it is constructing so `@types` seems good place

PR #98, opened 2020-03-31
Updated sdk to work with latest image CENNZnet 1.0.0

KarishmaBothara:


packages/types/src/extrinsic/types.ts
```diff
@@ -90,6 +91,129 @@ export interface SignatureOptions extends SignatureOptionsBase {
   doughnut?: AnyU8a | Doughnut;
   transactionPayment?: AnyU8a | ChargeTransactionPayment;
 }
+const AllHashers = {
```
benxgao: How about relocate these changes into separated files, so that we know they are types for metadata rather for extrinsic, easier for further refactoring and maintaining?

KarishmaBothara: sure.. will do that..

packages/types/src/extrinsic/types.ts
```diff
@@ -103,5 +227,23 @@ declare module '@polkadot/types/interfaceRegistry' {
     ChargeTransactionPayment;
     FeeExchange: FeeExchange;
     'Option<FeeExchange>': Option<FeeExchange>;
+    // v11
+    StorageHasherV11: StorageHasherV11;
+    DoubleMapTypeV11: DoubleMapTypeV11;
+    ExtrinsicMetadataV11: ExtrinsicMetadataV11;
+    // FunctionArgumentMetadataV11: 'FunctionArgumentMetadataV10';
```
holygits: Can remove?

packages/wallet/src/keyrings/HDKeyring.spec.ts
```diff
@@ -18,42 +18,42 @@ import {IKeyring} from '@plugnet/wallet/types';
...
-        beforeEach(async () => {
-            keyring = await HDKeyring.generate();
-            keyPairs = [];
-            for (let i = 0; i < times; i++) {
-                const keyPair = await keyring.addPair();
-                keyPairs.push(keyPair);
-            }
-            expect(keyPairs).toHaveLength(times);
-        });
+describe.skip('HDKeyring', () => {
```
benxgao: Does it mean wallet is not working? We may need a ticket accordingly and prioritise it later?

KarishmaBothara: we have a ticket for fixing wallet.. I will add a comment to the link to ticket?

packages/types/src/extrinsic/types.ts
KarishmaBothara: removed

packages/wallet/src/Wallet.spec.ts
```diff
@@ -37,8 +37,8 @@ const TEST_SIG_OPTS_V1 = {
   genesisHash: GENESIS_HASH,
   runtimeVersion: {} as any, //runtimeVersion is not used in v1
 };
-
-describe('a wallet', () => {
+// skipping the wallet related test - will fix with this task (https://centralitydev.atlassian.net/browse/IN-124)
```
benxgao: No need to add ticket link in source code:)

packages/wallet/src/keyrings/HDKeyring.spec.ts
benxgao: gotcha, thanks

Issue #84, opened 2020-01-21
failed on 'Type':: U8a


L-KH:
When I start to run examples, I get this error in example 6 to make a transfer: "api.js/docs/examples/promise/06_make_transfer/"
Please any solution to fix it?

`Unable to decode Vec on index 1 U8a: failed on 'Type':: U8a: failed on 'Type':: U8a: failed on 'Option':: U8a: failed on 'Type':: Invalid data passed to Mortal era
2020-01-21 12:30:19        RPC-CORE: getBlock (hash: Hash): SignedBlock:: createType(SignedBlock):: Struct: failed on 'block':: Struct: failed on 'extrinsics':: U8a: failed on 'Type':: U8a: failed on 'Type':: U8a: failed on 'Option':: U8a: failed on 'Type':: Invalid data passed to Mortal era
2020-01-21 12:30:19   API/DECORATOR: Error: FATAL: Unable to initialize the API: getBlock (hash: Hash): SignedBlock:: createType(SignedBlock):: Struct: failed on 'block':: Struct: failed on 'extrinsics':: U8a: failed on 'Type':: U8a: failed on 'Type':: U8a: failed on 'Option':: U8a: failed on 'Type':: Invalid data passed to Mortal era
    at EventEmitter.Init._onProviderConnect (C:\Users\user\node_modules\@polkadot\api\base\Init.js:88:23)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Error: Connection fail
    at EventEmitter.rejectError (C:\Users\user\node_modules\@cennznet\api\Api.js:49:16)
    at EventEmitter.emit (C:\Users\user\node_modules\eventemitter3\index.js:181:35)
    at Api.emit (C:\Users\user\node_modules\@polkadot\api\base\Events.js:25:31)
    at EventEmitter.Init._onProviderConnect (C:\Users\user\node_modules\@polkadot\api\base\Init.js:90:14)
    at process._tickCallback (internal/process/next_tick.js:68:7)`
![image](https://user-images.githubusercontent.com/12660772/72806969-a3babe00-3c56-11ea-8def-377e246f2bff.png)

Thanks

holygits:
should be resolved since #121

PR #100, opened 2020-04-02
Fix generate docs

KarishmaBothara:
So I am able to generate docs with 
`yarn run make-docs`
It deletes the old api_docs folder and creates new.
However this won’t resolve the issue https://github.com/cennznet/api.js/issues/84 as the examples in this is copied from polkadot and is not generated through script, just cross checked with Alex on this too..
https://github.com/cennznet/api.js/commit/f70580e1986f61966ccd4682865f56c0f80a1e5a#diff-e3e2a9bfd88566b05001b02a3f51d286 the commit history when it was copied from the polkadot in cennznet api.


benxgao:
Could we delete `pakcage-lock.json` file?

KarishmaBothara:
> Could we delete `pakcage-lock.json` file?

deleted

PR #103, opened 2020-04-21
Use bundled dependencies to fix polkadot version in consumer projects

KarishmaBothara:


packages/api/package.json
```diff
@@ -1,45 +1,53 @@
...
-    "typedoc": "^0.13.0",
-    "typedoc-plugin-markdown": "^1.1.19"
-  }
+    "name": "@cennznet/api",
+    "version": "1.0.0",
+    "description": "cennznet's javascript client",
+    "keywords": [
+        "CENNZNet"
+    ],
+    "author": "ian.he <ian.he@centrality.ai>",
```
alexsednz, line 8: Is the author still Ian? if not we can say "Centrality Developers".

packages/crml-attestation/package.json
```diff
@@ -13,12 +13,12 @@
     },
     "scripts": {},
     "dependencies": {
-        "@cennznet/api": "^1.0.0",
-        "@cennznet/types": "^1.0.0",
-        "@cennznet/util": "^1.0.0"
+        "@cennznet/api": "1.0.0",
```
alexsednz, line 16: Does it mean every time we release a new API, we should remember to manually update this line too?

alexsednz:
I don't feel comfortable with the general approach of this PR on fixing the dependencies to exact versions. I have recently been dealing with these dependency issues and could understand why you would have tried to get rid of the ranges that may induce uncertainties. However, the ranges when used correctly will let us remain updated (which means less security and non-security issues) and even when they fail, it's an important indication that some parts of our program may need to be updated too. However, as a short term workaround if we need to fix a version we should limit it to those specific ones that we need. Even in that case, we should try (~) instead of (^) first so if the breaking change is happening on major version changes we still keep updated with minor changes within the same major version number.

packages/api/package.json
```diff
@@ -1,45 +1,53 @@
...
+        "@cennznet/crml-attestation": "1.0.0",
+        "@cennznet/crml-cennzx-spot": "1.0.0",
+        "@cennznet/crml-generic-asset": "1.0.0",
+        "@polkadot/api-derive": "1.2.1",
+        "@cennznet/types": "1.0.0",
+        "@polkadot/api": "1.2.1",
+        "@polkadot/metadata": "1.2.1",
+        "@polkadot/rpc-core": "1.2.1",
+        "@polkadot/rpc-provider": "1.2.1",
+        "@polkadot/types": "1.2.1",
```
benxgao, line 31: Does `@cennznet/types@1.0.0` & its dependencies (including `@polkadot/types@1.2.1`) are bundled together?

If so can we assign a different version of `@polkadot/types` here?

packages/api/package.json
```diff
@@ -1,45 +1,53 @@
...
-    "@cennznet/wallet": "^0.20.7",
-    "@types/jest": "^23.3.5",
-    "jest": "24.1.0",
-    "rimraf": "^2.6.2",
-    "ts-mockito": "^2.3.1",
-    "typedoc": "^0.13.0",
-    "typedoc-plugin-markdown": "^1.1.19"
-  }
+    "name": "@cennznet/api",
+    "version": "1.0.0",
```
benxgao, line 3: As we discussed, we are going to implement semantic versioning after releasing `1.0.0`, and maintain a separated `changelog` for cennznet api. Are we going to override the previous release?

packages/api/package.json
benxgao: @holygits  What is the available email address for centrality developers?

ianhe8x:
Gonna say it will be problematic, since polkadot-api explicitly forbid multi-instance. If this lib bundled polkadot-api, then there will be problem when someone tries to use this lib in a project that already depends polkadot-api (even the versions of polkadot-api are the same).

PR #104, opened 2020-04-21
Fix Sylo Runtime Types

JCSanPedro:
When fetching values from substrate which had the types `PreKeyBundle` or `WithdrawnPreKeyBundle`, the values were decoding into empty bytes arrays, i.e `0x`. The types were simple, and removing the unnecessary constructor lines seemed to fix this issue.


holygits:
@JCSanPedro can you sign the commit then it's good to go

JCSanPedro:
@holygits done

PR #107, opened 2020-04-27
Release 1.0.1 (cennznet/cennznet:1.0.0)

benxgao:
Docker digest: sha256:c33d37c52a6b980ef7e93efbee7383a413616077ecfee9b7e9585d8f834793cd

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
   ],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "1.0.0"
+  "version": "1.0.1"
 }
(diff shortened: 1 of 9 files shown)


lerna.json
```diff
@@ -4,5 +4,5 @@
   ],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "1.0.0"
+  "version": "1.0.1"
```
alexsednz, line 7: Is that an intended change?

benxgao: It was changed by running `yarn run lerna version 1.0.1 --no-git-tag-version --no-push` which is suggested in Polkadot, rather than manually changed.

PR #112, opened 2020-05-25
Updated static metadata

KarishmaBothara:


benxgao:
Well... this PR is good enough for a quick and urgent patch, but a wider range of topics behind to be discussed:
- If we do want to include `staticMetadata` in the repo, we may need to consider clearly using semantic versioning, that is, we should bump a major version change in cennznet/api, whenever we upgraded metadata;
- It is not quite effective to store static metadata, as metadata could be changed whenever there is a new interface added (correct me if I am wrong); Dapps could cache metadata by themselves?
- Latest `polkadot` accepts string as metadata arg when api is initialised, but as we are depending on a old version of polkadot, we are not able to adopt that approach;
- We may need to upgrade CI/CD pipeline to use `matrix` to test multiple versions of APIs to be compatible with various versions of cennznet (That is fine if we keep dropping older versions of api and cennznet);

alexsednz:
> Well... this PR is good enough for a quick and urgent patch, but a wider range of topics behind it may need be discussed:
> 
> * If we do want to include `staticMetadata` in the repo, we may need to consider clearly using semantic versioning, that is, we may need to bump a major version in cennznet/api, whenever we upgraded metadata, due to compatible problem;
> * It is not quite effective to store static metadata, as metadata could be changed whenever there is a new interface added (correct me if I am wrong); Dapps could cache metadata by themselves?
> * Latest `polkadot` accepts string as metadata argument when api is initialised, but as we are depending on a old version of polkadot, we are not able to adopt that approach;
> * We may need to upgrade CI/CD pipeline to use `matrix` to test multiple versions of APIs to be compatible with various versions of cennznet (That is not a problem if we just want to keep dropping older versions of api and cennznet);

They seem good points to me. I guess the least we can do for some of them is to create tickets for either doing them later or researching them more. However, the first one seemed to me to be more urgent to address. @KarishmaBothara @holygits shouldn't we bump the version in cennznet/api?

packages/api/test/e2e/api.create.e2e.ts
```diff
@@ -59,4 +60,14 @@ describe('e2e api create', () => {
     await expect(Api.create({provider, timeout: 1})).rejects.toThrow(
       'Timed out in 1 ms.');
   });
+
+  it('Use custom metadata and test f it gets passed while creating API instance ', async () => {
+    const provider = 'wss://cennznet.unfrastructure.io/public/ws'; // Use Azalea
```
holygits, line 65: I think there's some jest config for this

KarishmaBothara: I couldn't use the initApiPromise from jest/initApiPromise.. cause this needs customise ApiOptions (with metadata)..

PR #117, opened 2020-07-08
Added support for liquidity RPC calls

KarishmaBothara:
Removed the old way to get liquidityPrice via derived query and update the existing test case to use rpc call.


packages/types/src/extrinsic/types.ts
```diff
@@ -105,5 +105,7 @@ declare module '@polkadot/types/interfaceRegistry' {
     FeeExchange: FeeExchange;
     'Option<FeeExchange>': Option<FeeExchange>;
     'Vec<(AssetId, AssetInfo)>': Vec<ITuple<[AssetId, AssetInfo]>>;
+    '(Balance, Balance)': '(Balance, Balance)';
```
holygits, line 108: This continues to confuse me 🤔

KarishmaBothara: Managed to fix this.. with any.. have removed it from types :)

packages/api/src/rpc/cennzx.ts
```diff
@@ -25,9 +25,25 @@ const sellPrice: RpcMethodOpt = {
   type: 'Balance',
 };
 
+const liquidityPrice: RpcMethodOpt = {
+  description: 'Retrieves the price of liquidity for a particular `asset_id`',
```
holygits, line 29: nit: "Get the price of liquidity for the given asset ID"

packages/api/src/rpc/cennzx.ts
```diff
@@ -25,9 +25,25 @@ const sellPrice: RpcMethodOpt = {
...
+  isOptional: true,
+  params: [createParam('AssetId', 'AssetId'), createParam('liquidityToBuy', 'Balance')],
+  type: '(Balance, Balance)' as any,
+};
+
+const liquidityValue: RpcMethodOpt = {
+  description: 'Retrieves the value of liquidity in the exchange for `asset_id` for `account`',
```
holygits, line 36: nit: Get the value of an account's liquidity for the given asset

PR #120, opened 2020-07-13
Chore/generate slim data

KarishmaBothara:
Added script to generate slim meta data for Sylo.. 
Meta data generated has been tested against connection and GA transfer.
The size of meta data has further reduced to 83,652.
Have tested it against local host, but can be modified once Azalea is upgraded?
`yarn generate-slim-metadata 'end-point'` will generate the new slim metadata

### DIFF
--- a/scripts/generateSlimMetadata.ts
+++ b/scripts/generateSlimMetadata.ts
@@ -1,0 +1,86 @@
+// Running this file with ~ node scripts/updateStaticMetadata.js will create a staticMetadata.ts on root directory which can used/copied
+import {Api} from '../packages/api/src/Api';
+import MetadataVersioned from '@polkadot/metadata/Metadata/MetadataVersioned';
+import {toCallsOnly} from '@polkadot/metadata/Metadata/util';
+
+import {createType} from '@cennznet/types';
+
+async function generateSlimMeta() {
+    // const provider = 'wss://cennznet.unfrastructure.io/public/ws'; // Use Azalea
+    const provider = 'ws://127.0.0.1:9944';
+    const api = await Api.create({provider});
+    const metadata = api.runtimeMetadata.toJSON();
+    const keep = [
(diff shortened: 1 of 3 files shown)


scripts/generateSlimMetadata.ts
```diff
@@ -0,0 +1,80 @@
+// Running this file with ~ yarn generate-slim-metadata will create slim metadata
+import {Api} from '../packages/api/src/Api';
+import MetadataVersioned from '@polkadot/metadata/Metadata/MetadataVersioned';
+
+import {createType} from '@cennznet/types';
+
```
holygits, line 6: would be nice to add some description of what it's removing etc.

scripts/generateSlimMetadata.ts
```diff
@@ -0,0 +1,80 @@
...
+async function generateSlimMeta() {
+    // const provider = 'wss://cennznet.unfrastructure.io/public/ws'; // Use Azalea
```
holygits, line 8: endpoint should be an option to pass in when run

scripts/generateSlimMetadata.ts
alexsednz: Or possibly what we are keeping, as it seems we are better aware of the parts that we are interested in.

KarishmaBothara: thanks people.. will update

scripts/generateSlimMetadata.ts
alexsednz: A comment explaining that when generating metadata we should try to connect to an endpoint which is running the desired version of cennznet.

scripts/generateSlimMetadata.ts
KarishmaBothara: made an attempt to describe.. let me know if it requires edition..

PR #121, opened 2020-07-13
Update and Fix examples and correct example names

alexsednz:
Some examples would be still failing but we will have separate tickets for them. It's just a step towards creating an improved getting started document.


docs/examples/promise/08_read_storage_at/index.js
```diff
@@ -11,7 +11,9 @@ const BOB = '5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty';
 const STAKING_ASSET = 16000;
```
holygits, line 11: could query `api.query.genericAsset.stakingAssetId()`

alexsednz: I will have that in mind if we go through improving these examples.

PR #122, opened 2020-07-14
Update get_started doc

alexsednz:
- Introduce examples
- Mention the initial steps
- Explain the required entries in package.json
- Walk the user through a very simple use case of cennznet/api
- Introduce Keyring


docs/GET_STARTED.md
```diff
@@ -1,6 +1,98 @@
...
+But if you are connecting to your own local node, it can be as simple as:
+```js
+const api = await Api.create();
+```
+Once the connection is created, you can use of one of the several api queries (located under api.query), RPCs (located under api.rpc) or Transactions (located under api.tx) to interact with the node. Here are examples of for each of these categories:
+
+Category | Example | Description
+--- | :---: | ---
+RPC | `api.rpc.system.chain()` | Get the change name
+Query | `api.query.genericAsset.freeBalance(CENNZ, Alice)` | How many CENNZs, Alice has in her free balance
```
trietnguyen267, line 56: I think we should declare CENNZ is a variable here and the real parameter input is the asset ID `16000` to avoid confusion

trietnguyen267:
just add a comment on method example params, LGTM!

docs/GET_STARTED.md
```diff
@@ -1,6 +1,98 @@
...
+```js
+const { Api } = require('@cennznet/api');
+```
+When you have the api, you can easily connect to any cennznet node by calling "connect". 
+The underlying mechanism to connect to a node is through websockets. 
+So what you need to have for this call is the websocket address of your node. 
+The cennznet's main network is residing at `wss://cennznet.unfrastructure.io/public/ws`. 
+If you use this address to connect to our main network which is called Azalea, you automatically get connected to one of the cennznet's validator/full nodes. 
+However you may want to run your own local node and connect to that for testing. To do so you can checkout the docker-compose script that we have provided in this project and also mentioned it later in this document. 
+Alternatively you can have a look at a https://github.com/cennznet/cennznet/wiki/Running-a-Full-Node.
```
KarishmaBothara, line 41: may be add a link for https://github.com/cennznet/cennznet/wiki/Running-a-Full-Node

docs/GET_STARTED.md
```diff
@@ -81,6 +81,17 @@ Now that you have added the key for Alice you can use it to sign an extrinsic:
 const extrinsic = api.tx.genericAsset.transfer(CENNZ, BOB, 12345);
 const hash = await extrinsic.signAndSend(alice);
 ```
+## FAQ
+> What is the difference between @cennznet/types and @polkadot/types?
```
holygits, line 85: perfect 👍

docs/GET_STARTED.md
```diff
@@ -1,6 +1,98 @@
...
+Query | `api.query.genericAsset.freeBalance(CENNZ, Alice)` | How many CENNZs, Alice has in her free balance
+Tx | `api.tx.genericAsset.transfer(CENNZ, BOB, 12345)` | Create an extrinsic for transferring 12345 CENNZs to Bob  
+
+## Keyring
+Some of the interactions with the node should be signed. 
+For example if you say Alice transfers 1 CENNZ to Bob, you should have the secret key of Alice and sign the created extrinsic before submitting it.
+More importantly if you use a secret, you don't want to expose it to anyone by using it directly in the code. 
+You may also have several keys for your different accounts. Where to keep those keys and how to use them? Keyring or wallet is our answer to this need. In order to use @cennznet/wallet in your project you should add the following dependency to your package.json:
+```json 
+"@cennznet/wallet": "^1.0.3"
```
KarishmaBothara, line 65: May be we don't want to mention about wallet, until it is working again

holygits: + recommend @polkadot/keyring for now

KarishmaBothara:
Looks really good.. added few comments..

docs/GET_STARTED.md
alexsednz: done

docs/GET_STARTED.md
alexsednz: done as a Note underneath the table.

holygits:
please squash merge feature branches otherwise history gets messy

PR #123, opened 2020-07-15
Updated azalea metadata

KarishmaBothara:
Updated static metadata for azalea


alexsednz:
Does this mean we now need to publish a new version of the API package again?

KarishmaBothara:
> Does this mean we now need to publish a new version of the API package again?

Not really.. It can wait, nothing will break... Sylo will use the slim metadata, so it won't impact them. Only anyone using the cennznet/api directly, it might slow down, cause it will fetch metadata and then use..

alexsednz:
Please do not merge as I have already cherry-picked your commits here to my branch. Thanks.

alexsednz:
Now we can safely close it as I merged the commits already.

PR #130, opened 2020-07-27
Add initialization to extrinsic phase

holygits:
Extrinsic phase on the node side has an additional variant which can create RPC errors decoding certain blocks.

- [x] add spec test

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -866,13 +866,6 @@
   dependencies:
     regenerator-runtime "^0.13.2"
 
-"@babel/runtime@^7.10.4":
-  version "7.10.5"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -899,6 +906,11 @@
     lodash "^4.17.13"
     to-fast-properties "^2.0.0"
 
+"@cennznet/cennznut-wasm@^0.0.1-dev.0":
```
KarishmaBothara, line 909: wondering why did it change the lock file.. as no dependency is modified..

holygits: Had other edits in this branch, will remove.

PR #133, opened 2020-07-29
Feat/doughnut extrinsic hash fix

KarishmaBothara:
Manage to fix the extrinsic hash which occurs with Doughnut.. Not sure what is the best way to fix.. This is one attempt..

Problem - Doughnut extends Raw (type) which says 
 * A basic wrapper around Uint8Array, with no frills and no fuss. It does differ
 * from other implementations where it will consume the full Uint8Array as passed to it.
 * As such it is meant to be subclassed where the wrapper takes care of the
 * actual lengths instead of used directly.

So when the extrinsic is sent from chain via get_block when it tries to decode Doughnut bytes, it behaves selfish and consumes all the following bytes (which were meant for era, nonce, transactionPayment and method).. so have added a static field which would set the length of doughnut when doughnut is attached to an extrinsic..let me know your thoughts..


holygits:
superseded by #134

PR #102, opened 2020-04-16
Refactor json PRC defination

benxgao:
Now that we have future plans on cennzx liquidity and potential api refactoring, I am on-holding any unnecessary code changes in api; I just made the minimum changes in this PR, and skipped cennzx e2e tests for now, which we could reenable once we implemented liquidity rpc.

This PR actually can be deprecated once we are able to upgrade to `polkadot@^1.8.1`, at that point, we could define RPC in type interfaces directly.

PR #138, opened 2020-08-02
Chore/129 dep updated

### LINKED ISSUE #129: Update to the latest polkadot-js packages
Polkadot API has experienced a major breaking change after the version 1.2.1 which has stopped our cennznet API to be able to update with the latest of this dependency. As the breaking change has happened a pretty long time ago and there have been several stable versions of Polkadot since that time, it is about time that we assessed how hard it is for us to adapt our own API to the latest.

alexsednz:
Closes #129 

Note 1: Test is in progress. It's an early reveal of the changes that we need to make for updating our API. The intention of making the PR at this stage is that the team get involved in the process and review at the earliest.

Note 2: I highly suggest that the reviewers go through the commits individually.


.editorconfig
```diff
@@ -3,9 +3,11 @@ root = true
 [*]
 indent_style = space
 indent_size = 2
+tab_width = 2
 end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
+max_line_length = 120
```
benxgao, line 10: May worth check existing effective Polkadot editor/eslint/prettier defined and effective in cennznet/api.js, which I suppose it is `80`, need to be confirmed.

package.json
```diff
@@ -19,38 +19,32 @@
...
+    "react-dom": "^16.13.1",
+    "react-is": "^16.13.1",
+    "semantic-release": "^15.8.0",
+    "styled-components": "^5.1.1",
+    "ts-jest": "^26.1.3",
+    "ts-node": "^8.10.2",
+    "tsconfig-paths": "^3.9.0",
+    "typedoc": "^0.17.8",
+    "typedoc-plugin-monorepo": "^0.1.0",
+    "typescript": "^3.9.7"
```
benxgao, line 47: I suppose the main purpose of this pull request is to particularly upgrade Pokadot dependencies, so that cennznet/api.js is able to be synchronised with  latest Pokadot dependencies.

IMO, to get the job be easier validated and tracked, it may be worth to try upgrade Polkadot related dependencies only in this PR; After that works, we could create a separated ticket/PR to upgrade non Polkadot related dependencies. On the other side, it could get reviewing job much easier and the PR could be done more effectively. What do you think?

.editorconfig
```diff
@@ -3,9 +3,11 @@ root = true
 [*]
 indent_style = space
 indent_size = 2
+tab_width = 2
```
benxgao, line 6: Could you please share any examples/styleguides/standards of popular Git repositories which are using `tab` defining indents, so that we could do more research and discussion? 
Otherwise, shall we keep `indent_size=2` as our default JavaScript indent size, and remove this config, which may trick editors which is extended with automatic formatting tools?

package.json
benxgao: As what we discussed in stand-up meeting, ignore this comment if it is already working in this PR. 

Otherwise, would reckon taking gradually upgrading packages in future PRs.

packages/api/src/rpc/cennzx.ts
```diff
@@ -1,49 +1,63 @@
...
+    description: 'Retrieves the spot exchange buy price',
+    params: [
+      {
+        name: 'AssetToBuy',
+        type: 'AssetId',
+      },
+      {
+        name: 'Amount',
+        type: 'Balance',
+      },
```
benxgao, line 12: It seems `AssetToPay` attribute is missing here;

Also there is a PR opening addressed relative refactoring where it shares the same approach like what are up to, for more details, please check out here - https://github.com/cennznet/api.js/pull/102/files

Relative refactoring may need to be conducted in `cennznet-ui`, and an opening PR addressed it as well, for more details, please check out here - https://github.com/cennznet/ui/pull/30/files

docs/examples/promise/06_make_transfer/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@cennznet/api": "^1.0.3",
-    "@cennznet/wallet": "^1.0.3"
+    "@plugnet/keyring": "^1.1.100"
```
KarishmaBothara, line 14: @polkadot/keyring

holygits:
Since the PR is about updating to latest polkadot, the wallet changes would be better done in another PR (it should also use **@polkadot/keyring** not **@plugnet**) it's making the review harder with all the changes not related to the update

packages/crml-cennzx-spot/src/CennzxSpotRx.ts
```diff
@@ -52,7 +47,6 @@ export class CennzxSpotRx {
     minLiquidity: AnyNumber,
     maxAssetAmount: AnyNumber,
     coreAmount: AnyNumber,
-    expire: AnyNumber
```
benxgao, line 55: Could we understand why `expire` is removed?

packages/crml-cennzx-spot/src/derives/assetWithdrawn.ts
```diff
@@ -13,48 +13,50 @@
 // limitations under the License.
 
 import {ApiInterfaceRx} from '@cennznet/api/types';
-import {AnyAssetId, AnyNumber} from '@cennznet/types/types';
+import {AnyNumber} from '@cennznet/types/types';
+import {AssetId} from '@cennznet/types';
```
benxgao, line 17: Could we understand the differences between '@cennznet/types/types' and '@cennznet/types', any chance to distinguish those two packages by names or even combined into one?

benxgao:
May need to merge/rebase latest `develop` branch sooner, due to release of https://github.com/cennznet/api.js/pull/140

alexsednz:
I'm going to close this PR as the branch "update/2.0.0" fully address the need for this update while re-using the polkadot codebase more efficiently.

PR #141, opened 2020-08-06
Add staking e2e tests

holygits:
Adds:
- Staking type definitions (RewardDestination)
- e2e tests for user and sudo staking public functions

Removes:
- Rewards type definitions

Changes:
- tx.e2e custom fee asset tests use a dedicated account for tests, not alice


packages/api/test/e2e/staking.e2e.ts
```diff
@@ -0,0 +1,304 @@
...
+import initApiPromise from '../../../../jest/initApiPromise';
+
+let api;
+const keyring = new Keyring({ type: 'sr25519' });
+let alice, bob;
+
+beforeAll(async () => {
+  await cryptoWaitReady();
+  api = await initApiPromise();
+  api.setSigner(keyring);
```
KarishmaBothara, line 31: I think this is not required

packages/api/test/e2e/staking.e2e.ts
```diff
@@ -0,0 +1,304 @@
...
+        expect(controller.unwrapOr(null)).toBeNull();
+        done();
+      }
+    });
+  });
+
+  /// Cancel enactment of a deferred slash. Can be called by root origin
+  /// passing the era and indices of the slashes for that era to kill.
+  it.skip('Cancel deferred slash', async () => { });
+
```
holygits, line 303: planning to look into the slashing stuff separately so I'll remove this from the PR

packages/api/test/e2e/staking.e2e.ts
```diff
@@ -0,0 +1,302 @@
...
+      (forcing: Forcing) => {
+        if (forcing.isForceAlways) done();
+      }
+    );
+    await api.tx.sudo.sudo(api.tx.staking.forceNewEraAlways()).signAndSend(alice);
+  });
+
+  test('Force unstake', async done => {
+    // Use charlie account as bob stash, it's simpler than funding a new account.
+    const bob_stash = keyring.addFromUri('//Charlie');
```
alexsednz, line 279: I would suggest you keep calling this key: charlie_stash. If it is okay for charlie to sign and send bob's bond tx, then the whole test reads right as well and there would be no confusion over bob and bob_stash.

alexsednz:
I just left one comment regarding the naming of an account in a test. Everything else LVGTM. Nice job.

packages/api/test/e2e/staking.e2e.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    await api.tx.staking.validate({ commission }).signAndSend(controller, checkCommission);
+  });
+
+  test('chill removes stash from validator candidacy', async done => {
+
+    const checkCommission = async ({ status }) => {
+      if (status.isInBlock) {
+        const prefs = ((await api.query.staking.validators(stash.address)) as ValidatorPrefs);
+        expect(prefs.commission.toNumber()).toEqual(0);
```
benxgao, line 164: Can we say that the validator is chilling if the commission number is 0?

packages/api/test/e2e/staking.e2e.ts
```diff
@@ -0,0 +1,302 @@
...
+        done();
+      };
+    };
+
+    await api.tx.staking.chill().signAndSend(controller, checkCommission);
+  });
+
+  test('setPayee changes reward destination', async done => {
```
benxgao, line 172: Is controller account by default a payee account, I suppose there is a `getPayee` method to grab it, right?

holygits: > Is controller account by default a payee account,
no, the payee account is originally set on a stash when bond() is called (it could be set to the stash itself or the controller)

> I suppose there is a getPayee method to grab it, right?
`api.query.staking.payee(<stash>)` is the getter method

packages/api/test/e2e/staking.e2e.ts
holygits: ended up finding the correct funded uri for bob/stash

packages/api/test/e2e/staking.e2e.ts
holygits: Yeah, it's the default value returned from storage in this case when nothing exists.
Someone could set their commission % to `0` intentionally and that could get confusing for UI checks...

PR #142, opened 2020-08-07
Made doughnut wasm packages as optional dependencies and used try/cat…

KarishmaBothara:
Don't know the best way of doing it.. For now made the doughnut wasm packages as optional dependencies and conditional used them in doughnut class to resolve issue on react-native for Sylo.
It wasn't possible with es6 import way to conditionally use a package..
Have tested it by removing the wasm packages from node modules and tested it


packages/api/package.json
```diff
@@ -33,8 +33,10 @@
   "resolution": {
     "@polkadot/keyring": "^1.6.1"
   },
+  "optionalDependencies": {
```
holygits, line 36: devDeps should be ok to leave as is

packages/api/test/e2e/doughnut.e2e.ts
```diff
@@ -23,7 +21,16 @@ import initApiPromise from '../../../../jest/initApiPromise';
 import { cryptoWaitReady } from '@plugnet/util-crypto';
 import Doughnut from '@cennznet/types/Doughnut';
 
+let DoughnutMaker, encodeCennznut;
 const [CENNZ, CPAY] = [16_000, 16_001];
+try {
```
holygits, line 26: using an `if` check should help bundlers ignore the import vs. the try catch

packages/api/test/e2e/doughnut.e2e.ts
```diff
@@ -13,8 +13,6 @@
 // limitations under the License.
 
 import { hexToU8a } from '@polkadot/util';
```
holygits, line 15: I don't think this needs to change, tests shouldn't break package build on normal install?

KarishmaBothara: ya... will revert this file

packages/types/src/Doughnut.ts
```diff
@@ -31,7 +30,16 @@ export default class Doughnut extends Raw {
     // NOT only doughnut bytes.
     // We must ensure that any doughnut specific bytes are consumed
     // so that the buffer [[value]] may be passed along to decode other types successfully
-    const doughnut = DoughnutCodec.decode(value as Uint8Array);
-    super(registry, doughnut.encode());
+    let doughnut;
+    // @ts-ignore
+    if (DOUGHNUT_SUPPORT === true) {
```
holygits, line 35: can you add a comment in this if block about the conditional compilation also

packages/types/src/Doughnut.ts
```diff
@@ -31,7 +30,17 @@ export default class Doughnut extends Raw {
...
+      // Added the above flag to support conditional compilation for exclusion/inclusion of wasm packages
+      const DoughnutCodec = require('@plugnet/doughnut-wasm');
+      doughnut = DoughnutCodec.Doughnut.decode(value as Uint8Array);
+      doughnut = doughnut.encode();
+    } else {
+      // fallback
+      doughnut = value;
```
alexsednz, line 42: I would prefer an error to be thrown, or the very least a log is generated under this condition.

holygits: fair point, maybe log and return empty Doughnut

PR #147, opened 2020-08-12
no coupling between session and account info query for staking

KarishmaBothara:
Modified the staking query to be independent of each other. This works well on UI

### DIFF
--- a/packages/api/src/derives/staking/index.ts
+++ b/packages/api/src/derives/staking/index.ts
@@ -1,6 +1,6 @@
 export {
-  queryStakingAccountDetails as accountInfo,
-  querySession as sessionDetails,
+  queryStakingAccountInfo as accountInfo,
+  querySessionInfo as sessionKeyInfo,
   DerivedStakingInfo,
   DerivedSessionKeyInfo,
 } from './stakingAccount';
--- a/packages/api/src/derives/staking/index.ts
+++ b/packages/api/src/derives/staking/index.ts
@@ -1,1 +1,11 @@
-export {queryStakingAccountDetails as accountInfo, querySession as sessionDetails} from './stakingAccount';
+export {
+  queryStakingAccountDetails as accountInfo,
+  querySession as sessionDetails,
+  DerivedStakingInfo,
+  DerivedSessionKeyInfo,
+} from './stakingAccount';
+export * from '@polkadot/api-derive/staking/account';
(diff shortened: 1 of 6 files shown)


packages/api/src/derives/staking/index.ts
```diff
@@ -1 +1,11 @@
...
+  queryStakingAccountDetails as accountInfo,
+  querySession as sessionDetails,
+  DerivedStakingInfo,
+  DerivedSessionKeyInfo,
+} from './stakingAccount';
+export * from '@polkadot/api-derive/staking/account';
+export * from '@polkadot/api-derive/staking/controllers';
+export * from '@polkadot/api-derive/staking/electedInfo';
+export * from '@polkadot/api-derive/staking/overview';
+export * from '@polkadot/api-derive/staking/validators';
```
benxgao, line 11: Could we have a general practice guideline on how we are dealing with exporting `polkadot/api` in `cennznet/api`, like whether we encoverage consumers considering `cennznet/api` as the only api interface entry point, or suggest them use polkadot interfaces if possible where we only include cennznet customized api? @holygits

packages/api/src/derives/staking/index.ts
```diff
@@ -1 +1,11 @@
-export {queryStakingAccountDetails as accountInfo, querySession as sessionDetails} from './stakingAccount';
+export {
+  queryStakingAccountDetails as accountInfo,
+  querySession as sessionDetails,
```
holygits, line 3: Nit: a bit more consistency with naming would be good here
```suggestion
  queryStakingAccountInfo as accountInfo,
  querySessionInfo as sessionKeyInfo,
```

packages/api/src/derives/staking/index.ts
holygits: @cennznet/api, users won't import @polkadot/api directly

packages/api/src/derives/staking/index.ts
KarishmaBothara: done

packages/api/src/derives/staking/index.ts
benxgao: Ok, then lgtm

KarishmaBothara:
> looks good in general.
> Thinking we should move the derived session query to `api.dervie.session.keyInfo` for better discoverability/consistency

Made these changes

PR #151, opened 2020-08-16
Release/1.2.0

KarishmaBothara:
Minor version release after supporting new api derive interfaces

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
   ],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "1.1.2"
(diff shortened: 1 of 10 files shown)


KarishmaBothara:
> I think the convention so far is to bump every package in the monorepo together.
> Simplifies compatibility as we don't intend these to be mixed-and-matched.
> Everything should go to `1.2.0`

Ya.. I forgot to run the command - yarn run lerna version 1.2.0 --yes --no-git-tag-version --no-push again.. I used it earlier with 1.1.3.. 
Made these changes

PR #159, opened 2020-08-24
Chore/convert tslint to eslint

KarishmaBothara:
Steps:
1. Converted existing tslint to eslint using https://github.com/typescript-eslint/tslint-to-eslint-config
2.  Removed tslint.json
3.  Updated package.json to use eslint linter
4.  Commented some rules from .eslintrc for now

Steps ahead:
Plan to create PR against this PR for every rule commented in eslintrc as with those rules I have around 648 errors.


.eslintrc.js
```diff
@@ -0,0 +1,61 @@
...
+    "parser": "@typescript-eslint/parser",
+    "parserOptions": {
+        "project": "tsconfig.json",
+        "sourceType": "module"
+    },
+    "plugins": [
+        "@typescript-eslint"
+    ],
+    "rules": {
+        "@typescript-eslint/await-thenable": "error",
```
holygits, line 27: how different is this config from `@typescript-eslint/recommended`?
wondering if it's better to start from a basic config than this as it's hard to say which lint to keep or remove

package.json
```diff
@@ -62,7 +62,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged --pattern 'packages/**/src/**/*' && npx tslint -c tslint.json --project ."
+      "pre-commit": "pretty-quick --staged --pattern 'packages/**/src/**/*' && npx eslint . --ext .ts,.tsx"
```
holygits, line 65: I don't think we have `.tsx` in this project, also it should be same as `npm run lint` https://github.com/cennznet/api.js/pull/159/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R19

.eslintrc.js
KarishmaBothara: https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin has all the rules
@typescript-eslint/recommended which is based on syntax (no type checking) and
@typescript-eslint/recommended-requiring-type-checking which are based on semantics (type-checking)
 It is a superset of the rules we have in this config.. this config is what we had as tslint..

KarishmaBothara: Among all the typescript lint rules - the following we have which do not belong to recommended category (Key: :heavy_check_mark: = recommended)
@typescript-eslint/explicit-member-accessibility  @typescript-eslint/promise-function-async @typescript-eslint/member-ordering  @typescript-eslint/naming-convention @typescript-eslint/dot-notation 
There are several others - which appears as general rule not ts specific
"complexity": "error",
"curly": [
    "error",
    "multi-line"
],
"default-case": "error",
"eqeqeq": [
    "error",
    "always"
],
"import/no-extraneous-dependencies": "off",
"import/order": "error",
"no-console": "error",
"no-duplicate-imports": "error",
"no-magic-numbers": "error",
"no-return-await": "error",
"no-undef-init": "error",
"prefer-const": "error",
"prefer-template": "error",
"use-isnan": "error"

package.json
KarishmaBothara: done

.eslintrc.js
alexsednz: Shouldn't there be an extend command like the following as mentioned https://github.com/eslint/eslint/tree/master/packages/eslint-config-eslint:
{
    "extends": "eslint"
}

KarishmaBothara: I am adding that as a separate PR - https://github.com/cennznet/api.js/pull/160 
Have extended the support for "plugin:@typescript-eslint/recommended" and in the process to extend
eslint:recommended and plugin:@typescript-eslint/recommended-requiring-type-checking in the same PR

benxgao (approved):
Checked locally by running `yarn lint`, without linting errors. Looks good to me.

alexsednz (approved):

.eslintignore
```diff
@@ -0,0 +1,7 @@
+/build/**
+/api_docs/**
+**/node_modules/**
+**/test/**/*
```
benxgao, line 4: Shall we take test files into account as well?

package.json
```diff
@@ -62,7 +67,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged --pattern 'packages/**/src/**/*' && npx tslint -c tslint.json --project ."
+      "pre-commit": "pretty-quick --staged --pattern 'packages/**/src/**/*' && npx eslint . --ext .ts"
```
benxgao, line 70: Shall we add `js` and `json` in `ext` as well, in case of any js script/config files like `eslint.json` or `xxx.config.js`

.eslintignore
KarishmaBothara: Enabling it for test adds additional 101 errors.. we can create a ticket later to address those and remove test from ignore list.. Right now I have kept it how it was with tslint.json were test files were ignored..

benxgao: Gotcha, that makes sense. Would you mind creating a ticket or a reminder somewhere so that we could revisit it later?

package.json
KarishmaBothara: I believe it won't help to check against json files.. cause syntax and type checking won't be required in case of json.. js can be added... created a spike ticket for having js as extension.. Tried locally to run.. but the command  npx eslint . --ext .ts,.js kept running for last 30mins..

.eslintignore
KarishmaBothara: done

package.json
benxgao: Good to know that, thanks.

.eslintrc.js
```diff
@@ -0,0 +1,86 @@
...
+        '@typescript-eslint/no-unsafe-call': "off",
+        '@typescript-eslint/no-unsafe-member-access': "off",
+        '@typescript-eslint/no-unsafe-return': "off",
+        '@typescript-eslint/no-unnecessary-type-assertion': "off",
+        '@typescript-eslint/restrict-plus-operands': "off",
+        '@typescript-eslint/restrict-template-expressions': "off",
+        // Support recommended-requiring-type-checking (specific to this project)
+        '@typescript-eslint/unbound-method': "warn",
+        '@typescript-eslint/no-floating-promises': "warn",
+        // Making affected recommended types off for now
```
holygits, line 45: comment is out of date

package.json
```diff
@@ -16,21 +16,25 @@
     "make-docs": "rm -rf api_docs && typedoc --plugin typedoc-plugin-markdown --options typedoc.config.js .",
     "metadata-docs": "node ./scripts/MetadataMdWrapper.js",
     "generate-slim-metadata": "node ./scripts/generateSlimMetadataWrapper.js",
-    "lint": "npx tslint -c tslint.json --project . && npx prettier --list-different './packages/**/src/*.ts'"
+    "lint": "eslint -c .eslintrc.js --ext .ts"
```
holygits, line 19: should this still be `npx eslint` otherwise it requires global install?

holygits: Ok it works using local deps.
turns out PATH will be extended to include: `./node_modules/.bin/eslint`

PR #163, opened 2020-08-25
Fix/disable no magic number rule

KarishmaBothara:
PR - Fix or disables "no magic number" rule..
Some places it is just checking for 0.. Have disabled the check their over creating a variable.

### DIFF
--- a/packages/api/src/derives/fees/estimateFee.ts
+++ b/packages/api/src/derives/fees/estimateFee.ts
@@ -29,11 +29,12 @@ import {catchError, first, map, switchMap} from 'rxjs/operators';
 export function estimateFee(api: ApiInterfaceRx) {
   // We generate fake signature data here to ensure the estimated fee will correctly match the fee paid when the extrinsic is signed by a user.
   // This is because fees are currently based on the byte length of the extrinsic
+  const genesisBlock = 0;
   return ({extrinsic, userFeeAssetId, maxPayment}): Observable<any> => {
     return combineLatest([
       api.rpc.state.getRuntimeVersion(),
       api.rpc.chain.getBlockHash(),
-      api.rpc.chain.getBlockHash(0),
+      api.rpc.chain.getBlockHash(genesisBlock),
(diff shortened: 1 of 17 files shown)


benxgao:
Now that we had prevented linting errors from `yarn lint`, how about we pause turning on them one by one for now.

We could have a session to discuss within the team to decide which rules we should keep and which to remove. After that, we could split any further relevant refatoring tasks within the team.

Then we could close this ticket https://centralitydev.atlassian.net/jira/software/projects/IND/boards/142?selectedIssue=IND-239 , and open a new one for further discussion. What do you think? cc @alexsednz @holygits

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -29,11 +29,12 @@ import {catchError, first, map, switchMap} from 'rxjs/operators';
 export function estimateFee(api: ApiInterfaceRx) {
   // We generate fake signature data here to ensure the estimated fee will correctly match the fee paid when the extrinsic is signed by a user.
   // This is because fees are currently based on the byte length of the extrinsic
+  const genesisBlock = 0;
```
holygits, line 32: `genesisBlockNumber`

packages/types/src/runtime/sylo/index.ts
```diff
@@ -85,8 +86,9 @@ export class PendingInvite extends Struct {
   }
 
   toJSON() {
+    const bitLenght = -1;
```
holygits, line 89: `bitLength`

holygits (commented):
This lint might be better as a warning, especially don't like we need to annotate lot's of obvious zero checks

KarishmaBothara:
> This lint might be better as a warning, especially don't like that we'd need to annotate lots of obvious zero checks

In that case.. shall I decline this PR? and just turn it off to warning.

KarishmaBothara:
Not required. Planning to turn this rule to warning

PR #172, opened 2020-09-18
Support nodes develop branch

KarishmaBothara:
Made changes to support the latest CENNZnet node's develop branch.
 Updated the balance where it was big value to smaller value
Added a new test case to updated payee as any account
Updated metadata
One thing that I could not do is to update the interfaceRegistry to reflect the new type definition for RewardDestination..module augmentation used does not let me do that.. any pointers appreciated..


KarishmaBothara:
> Made changes to support the latest CENNZnet node's develop branch.
> Updated the balance where it was big value to smaller value
> Added a new test case to updated payee as any account
> Updated metadata
> One thing that I could not do is to update the interfaceRegistry to reflect the new type definition for RewardDestination..module augmentation used does not let me do that.. any pointers appreciated..

Type definition for RewardDestination is now merged to support 'account' type.. thanks @holygits for this..

packages/api/test/e2e/cennzx.e2e.ts
```diff
@@ -25,7 +25,7 @@ describe('CENNZX e2e queries/transactions', () => {
   describe('Queries()', () => {
 
     it("Deposit liquidity in CENNZ asset's pool", async done => {
-        const amount = 30000000000000;
+        const amount = 3000000;
```
holygits, line 28: 3_000_000

packages/types/src/runtime/staking/types.spec.ts
```diff
@@ -15,5 +15,17 @@ describe('RewardDestination', (): void => {
...
+
+    const destinationAccountAddress = '5FEe8Ht1ZTzNjQcvrxbLxnykA2EXfqN5LMog2gaNPus4tfZR';
+    const destinationAccount = createType(registry, 'RewardDestination', {account: destinationAccountAddress}, 2);
+    expect(destinationAccount.asAccount.toString()).toEqual(destinationAccountAddress);
+    expect(destinationAccount.isAccount).toBeTruthy();
+    const destinationAccountStr = createType(registry, 'RewardDestination', {account: destinationAccountAddress});
+    expect(destinationAccountStr.asAccount.toString()).toEqual(destinationAccountAddress);
+    expect(destinationAccountStr.isAccount).toBeTruthy();
+
+    const destinationPayee = new RewardDestination(registry, {account: destinationAccountAddress});
```
holygits, line 27: nit:
```suggestion
    const destinationPayee = new RewardDestination(registry, { account: destinationAccountAddress });
```

packages/types/src/runtime/staking/index.ts
```diff
@@ -13,11 +13,30 @@
...
 // Specifies which account staking rewards should be paid too.
 // Note: Polkadot has an additional 'Staked' option which is not supported in CENNZnet
 // due to the fact that the reward currency is different from the staked currency.
-export class RewardDestination extends Enum.with({
-  Stash: 'Null',
-  Controller: 'Null',
-}) {}
+export class RewardDestination extends Enum {
+  constructor(registry: Registry, value: any) {
+    super(registry, {Stash: 'Null', Controller: 'Null', Account: 'AccountId'}, value);
```
holygits, line 24: nit:
```suggestion
    super(registry, { Stash: 'Null', Controller: 'Null', Account: 'AccountId' }, value);
```

holygits (commented):
changes lgtm!

holygits (approved):

packages/types/src/runtime/staking/types.spec.ts
holygits: I though eslint would do this so not sure why it's different

KarishmaBothara: I will search which rule will do this and enable it explicitly in the eslintrc..

KarishmaBothara: "object-curly-spacing": [2, "always"] -- if I add this rule, I have 808 errors.. I will update this PR without the rule and modify at these two place and have another PR to fix those 808 errors.. there as option to fix it with code..will use that..

PR #180, opened 2020-09-22
Make attestation derive only

holygits:
Remove crml-generic-asset package
Attestation is no longer a package, it's files and tests are moved to packages/api/src/derives/attestation


typedoc.config.js
```diff
@@ -21,7 +19,7 @@ module.exports = {
     ["@cennznet/api", "(.*packages\/api\/.*|node_modules\/@cennznet\/api\/.*)", "packages/api/README.md"],
     ["@cennznet/types", "(.*packages\/types\/.*|node_modules\/@cennznet\/types\/.*)", "packages/types/README.md"],
     ["@cennznet/util", "(.*packages\/util\/.*|node_modules\/@cennznet\/util\/.*)", "packages/util/README.md"],
-    ["@cennznet/crml-generic-asset", "(.*packages\/crml-generic-asset\/.*|node_modules\/@cennznet\/crml-generic-asset\/.*)", "packages/crml-generic-asset/README.md"],
+    ["@cennznet/crml-cennzx-spot", "(.*packages\/crml-attestaation\/.*|node_modules\/@cennznet\/crml-attestaation\/.*)", "packages/crml-attestation/README.md"],
```
KarishmaBothara, line 22: "@cennznet/crml-cennzx-spot" -> "@cennznet/attestation" and attestaation -> attestation

jest/initApiPromise.js
```diff
@@ -1,14 +1,12 @@
...
 import { Api as ApiPromise } from '../packages/api/src/Api';
 
 const initApiPromise = async () => {
   const providerUrl = config.wsProvider[`${process.env.TEST_TYPE}`] || 'ws://localhost:9944';
-  const wsProvider = new WsProvider(providerUrl);
   console.log('providerUrl', providerUrl);
   console.log('process.env.TEST_TYPE: ', process.env.TEST_TYPE);
 
-  const api = await ApiPromise.create({ provider: wsProvider });
+  const api = await ApiPromise.create(providerUrl);
```
KarishmaBothara, line 9: I think we need to pass it as an object.. i.e ApiOption so it has to be ({provider: providerUrl})

packages/api/src/derives/attestation/attestationRx.e2e.ts
```diff
@@ -0,0 +1,261 @@
...
+import { filter, first } from 'rxjs/operators';
+
+import { ApiRx } from '@cennznet/api';
+import { AttestationValue, TypeRegistry } from '@cennznet/types';
+import testKeyring from '@polkadot/keyring/testing';
+import { KeyringPair } from '@polkadot/keyring/types';
+import { cryptoWaitReady } from '@polkadot/util-crypto';
+
+import initApiRx from '../../../../../jest/initApiRx';
+import { H256 } from '@cennznet/types';
```
holygits, line 24: should be grouped with the other `@cennznet/types` imports (shouldn't we have lint for this 🤔)

packages/api/src/derives/attestation/attestationRx.e2e.ts
```diff
@@ -0,0 +1,261 @@
...
+            }
+          }
+        });
+    });
+  });
+
+  describe('Get Mutiple Claims Claim', () => {
+    it('should get a claim with a specific issuer and holder', done => {
+      api.derive.attestation
+        .getClaims(holder.address, [issuer.address, issuer2.address], [topic, topic2])
```
holygits, line 213: Isn't it possible that this test does nothing and still passes (false positive) i.e. if doesn't go into the `if` blocks?
I wonder if I may've broken the API here as it used to pass ok I guess?
`.subscribe(({claims}) => {`
either way this test needs to be reworked

packages/api/src/derives/attestation/attestationRx.e2e.ts
```diff
@@ -0,0 +1,261 @@
+// Copyright 2019 Centrality Investments Limited
```
holygits, line 1: Copyright 2019-2020

packages/api/src/derives/attestation/attestationRx.e2e.ts
KarishmaBothara: https://eslint.org/docs/rules/ the recommended does not include - sort-imports... we can add it specifically as a rule in the config.

KarishmaBothara: cause it is included in test and we have skipped all **/*.e2e.ts

KarishmaBothara: made this change

packages/api/src/derives/attestation/attestationRx.e2e.ts
KarishmaBothara: This can happen sometime when the extrinsic is not yet applied and subscribe would give value before and also once extrinsic is submitted and storage is changed.. I have updated the test..

alexsednz (approved):

PR #184, opened 2020-10-01
Chore/update readme

KarishmaBothara:
Updated read me inside packages/api as well as in some docs example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,20 @@
 
 The CENNZnet JavaScript API library for browsers and Node.js.
 
+## Quick Start
+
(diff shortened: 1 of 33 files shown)


holygits:
Thanks for this! could you add this section to the main README which will show on github and the NPM landing pages to make it very obvious :D 

---

## Quick Start
You **must** use `yarn` with @cennznet/api
and set the following resolutions in your `package.json`
```
<resolutions>
```
Otherwise your install may break due to breaking changes in downstream package versions.

holygits (approved):

PR #190, opened 2020-10-16
Fix rxapi.queries.e2e.ts

alexsednz:
Eventemitter was internal to ProviderInterface and programming towards the internal was wrong in the first place. Later Polkadot changed that internal private part and introduced "on" method for handling the error.


KarishmaBothara (approved):
lgtm

PR #192, opened 2020-10-19
Fix/rpc queries

alexsednz:
Let's merge this as it already fixes the cennzx and generic asset rpc queries and jest debugging.


packages/api/src/rpc/index.ts
```diff
@@ -1,9 +1,11 @@
...
-//
-// export default interfaces;
+import cennzxRpc from './cennzx';
+import genericAssetRpc from './genericAsset';
+import paymentRpc from './payment';
+
+const interfaces = {
+  cennzx: cennzxRpc,
+  genericAsset: genericAssetRpc,
+  payment: paymentRpc,
```
holygits, line 8: nit: can we rely on an `@polkadot` provided definition for `payment`?

holygits (approved):

packages/api/src/rpc/payment.ts
```diff
@@ -0,0 +1,18 @@
+const paymentRpc = {
```
KarishmaBothara, line 1: Just one doubt.. Why do we need to have paymentRpc call, isn't it already available with polkadot/api by default?

packages/api/src/rpc/cennzx.ts
```diff
@@ -1,51 +1,63 @@
...
-// export default [
-//   { ...createMethod(section, 'buyPrice', buyPrice), name: 'buyPrice' },
-//   { ...createMethod(section, 'sellPrice', sellPrice), name: 'sellPrice' },
-//   { ...createMethod(section, 'liquidityPrice', liquidityPrice), name: 'liquidityPrice' },
-//   { ...createMethod(section, 'liquidityValue', liquidityValue), name: 'liquidityValue' },
-// ];
+const cennzxRpc = {
+  buyPrice: {
+    description: 'Retrieves the spot exchange buy price',
+    params: [
```
KarishmaBothara, line 4: we would need one more param right?
AssetToPay?

packages/api/src/rpc/payment.ts
alexsednz: It should be. I say we recently enabled the payment rpc on cennznet: update/2.0.0 and thought we had to add it like a custom rpc to our api. I will remove it.

packages/api/src/rpc/cennzx.ts
alexsednz: Yup, it had changed since the last time we made this json. Thank you for noticing it.

packages/api/src/rpc/index.ts
alexsednz: Yup true, we should be able. I will investigate why queryInfo is then not part of the payment rpc coming from Polkadot.

packages/api/src/rpc/cennzx.ts
```diff
@@ -1,51 +1,67 @@
...
+      {
+        name: 'AssetToSell',
+        type: 'AssetId',
+      },
+      {
+        name: 'Amount',
+        type: 'Balance',
+      },
+      {
+        name: 'AssetToPay',
```
KarishmaBothara, line 32: I think name 'AssetToPayout' seems more meaningful as it will appear in UI..
#[rpc(name = "cennzx_sellPrice")]
	// TODO: change to Result<Balance> once https://github.com/serde-rs/serde/pull/1679 is merged
	fn sell_price(&self, asset_to_sell: AssetId, amount_to_buy: Balance, asset_to_payout: AssetId) -> Result<u64>;

alexsednz: Done

KarishmaBothara (approved):
lgtm

PR #193, opened 2020-10-19
Update/2.0.0 fix generate slim metadata

KarishmaBothara:
Updated the script to generate slim metadata.. Have tested the metadata with API creation.. works well..

### DIFF
--- a/scripts/generateSlimMetadata.ts
+++ b/scripts/generateSlimMetadata.ts
@@ -26,7 +26,7 @@ async function generateSlimMeta() {
     'SyloInbox',
     'SyloResponse',
     'SyloVault',
-    'CennzxSpot',
+    'Cennzx',
   ];
   let magicNumber = metadata.magicNumber;
   let modules = api.runtimeMetadata.asLatest.modules;
--- a/scripts/generateSlimMetadata.ts
+++ b/scripts/generateSlimMetadata.ts
@@ -35,8 +35,8 @@ async function generateSlimMeta() {
         if (KEEP.indexOf(m.name.toJSON()) >= 0) {
             const module = m.toJSON();
             removeKeys(module, "documentation");
-            // Creating ModuleMetadataV11
-            let modifiedModule = api.registry.createType( 'ModuleMetadataV11', module);
+            // Creating ModuleMetadataLatest
+            let modifiedModule = api.registry.createType( 'ModuleMetadataLatest', module);
             newModules.push(modifiedModule);
         } else {
             // Push an empty module
--- a/scripts/generateSlimMetadata.ts
+++ b/scripts/generateSlimMetadata.ts
@@ -1,88 +1,86 @@
-// // Running this file with ~ yarn generate-slim-metadata 'wss://cennznet.unfrastructure.io/public/ws' will create slim metadata
-// import {Api} from '../packages/api/src/Api';
-// import MetadataVersioned from '@polkadot/metadata/Metadata/MetadataVersioned';
-//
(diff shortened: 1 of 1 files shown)


holygits (approved):
lgtm.
small nit is it's missing some formatting 
adding a unit test to connect with this metadata should be straightforward? could be a nice improvement for the future

KarishmaBothara:
> lgtm.
> small nit is it's missing some formatting
> adding a unit test to connect with this metadata should be straightforward? could be a nice improvement for the future

Reformatted with prettier.. We have an integration test to check if a meta data option is passed, will it be used and connected fine.. ya but we can have a unit test too in future..

scripts/generateSlimMetadata.ts
```diff
@@ -1,88 +1,85 @@
...
+    'Timestamp',
+    'TransactionPayment',
+    'GenericAsset',
+    'SyloGroups',
+    'SyloE2EE',
+    'SyloDevice',
+    'SyloInbox',
+    'SyloResponse',
+    'SyloVault',
+    'CennzxSpot',
```
alexsednz, line 29: Is it still called CennzxSpot or simply Cennzx?

scripts/generateSlimMetadata.ts
```diff
@@ -1,88 +1,85 @@
...
+    } else {
+      // Push an empty module
+      newModules.push({ name: m.name, calls: m.calls, events: m.events });
+    }
+  }
+  let extrinsic = metadata.asLatest.extrinsic;
+  let filteredModule = api.registry.createType('Vec<ModuleMetadataLatest>', newModules);
+  let filteredMetaLatest = api.registry.createType('MetadataLatest', { modules: filteredModule, extrinsic });
+  const filteredMetadataAll = api.registry.createType('MetadataAll', filteredMetaLatest, 12);
+  const mVersionedNew = new MetadataVersioned(api.registry, {
```
alexsednz, line 50: Is 12 or 11 the length of KEEP? If so, can you please replace it with KEEP.len?

scripts/generateSlimMetadata.ts
```diff
@@ -1,88 +1,85 @@
...
+      newModules.push(modifiedModule);
+    } else {
+      // Push an empty module
+      newModules.push({ name: m.name, calls: m.calls, events: m.events });
+    }
+  }
+  let extrinsic = metadata.asLatest.extrinsic;
+  let filteredModule = api.registry.createType('Vec<ModuleMetadataLatest>', newModules);
+  let filteredMetaLatest = api.registry.createType('MetadataLatest', { modules: filteredModule, extrinsic });
+  const filteredMetadataAll = api.registry.createType('MetadataAll', filteredMetaLatest, 12);
```
alexsednz, line 49: Is 12 or 11 the length of KEEP? If so, can you please replace it with KEEP.len?

scripts/generateSlimMetadata.ts
KarishmaBothara: that's right.. we modified it.. thanks

scripts/generateSlimMetadata.ts
KarishmaBothara: 11 earlier was the index we were at earlier.. now its 12.. I will trying to search for some generic way like 'latest'.. but unfortunately could not find any..

KarishmaBothara: no..its not length.. its the ENUM index..

alexsednz (approved):

PR #196, opened 2020-10-22
Chore/final cleanup

KarishmaBothara:
Cleaned up few things and tried to solve as many lint error as possible before release branch work

### DIFF
--- a/packages/api/src/Api.ts
+++ b/packages/api/src/Api.ts
@@ -29,7 +29,7 @@ export class Api extends ApiPromise {
     const api = new Api(options);
     return withTimeout(
       new Promise((resolve, reject) => {
-        const rejectError = err => {
+        const rejectError = () => {
           // Disconnect provider if API initialization fails
           api.disconnect();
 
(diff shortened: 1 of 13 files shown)


packages/api/src/Api.ts
```diff
@@ -29,7 +29,7 @@ export class Api extends ApiPromise {
     const api = new Api(options);
     return withTimeout(
       new Promise((resolve, reject) => {
-        const rejectError = err => {
+        const rejectError = () => {
```
KarishmaBothara, line 32: we don't use the err object and since was cause lint error

holygits (approved):
lgtm, no concerns

packages/api/src/derives/cennzx/assetWithdrawn.ts
```diff
@@ -36,9 +36,9 @@ export function assetToWithdraw(instanceId: string, api: ApiInterfaceRx) {
       map(([tradeAssetReserve, coreAssetReserve, totalLiquidity]) =>
         getAssetToWithdraw(
           new BN(liquidity.toString()),
-          coreAssetReserve as any,
-          tradeAssetReserve as any,
-          totalLiquidity as any
+          new BN(coreAssetReserve.toString()),
```
alexsednz, line 39: Are toString calls here necessary?

KarishmaBothara: yes.. because the coreAssetReserve we receive is of type Balance and we need to string of balance to create BN object..

alexsednz (approved):

PR #197, opened 2020-10-22
Maintain the behaviour of ApiRx connection establishment

alexsednz:
Maintain the behaviour strictly as it used to be. During connection setup, there is an opportunity window for a disconnect to be useful between when a connection is made successfully and when the API is "ready".


KarishmaBothara (approved):
lgtm

PR #200, opened 2020-10-26
1.3.0 alpha.1

KarishmaBothara:
Releasing the alpha version

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -18,7 +18,7 @@
         "test": "echo \"Error: run tests from root\" && exit 1"
     },
     "dependencies": {
-        "@cennznet/util": "1.3.0-alpha.1",
+        "@cennznet/util": "^1.3.0-alpha.1",
         "@polkadot/keyring": "^3.6.1",
         "@polkadot/types": "^2.3.1",
         "@polkadot/util": "^3.5.1",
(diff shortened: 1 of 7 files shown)


alexsednz (approved):

packages/types/package.json
```diff
@@ -18,7 +18,7 @@
         "test": "echo \"Error: run tests from root\" && exit 1"
     },
     "dependencies": {
-        "@cennznet/util": "^1.2.1",
+        "@cennznet/util": "1.3.0-alpha.1",
```
alexsednz, line 21: This should be ^1.3.0 or remain ^1.2.1 until we have published cennzet/util. Isn't it?

alexsednz: The right order of publishing according to the hierarchy of our packages seem to be util -> types -> api, can we enforce that it lerna.json?

KarishmaBothara: no.. how monorepo works is.. It will use 1.3.0-alpha.1 from the current packages where packages/util has version as 1.3.0-alpha.1.. Does not have to wait till one package is published and then we can publish the dependent one..
Monorepo will publish all at once..

alexsednz: Would ^1.3.0 work as well? If so it would be preferred.

PR #208, opened 2020-11-05
Updated API to support 1.2.0-rc1-node

KarishmaBothara:
Testing the image cennznet/cennznet:1.2.0-rc1 with API


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3'
 services:
   testnet_node_alice:
     container_name: testnet_node_alice
-    image: cennznet/cennznet:develop-e8f05b
+    image: cennznet/cennznet:1.2.0-rc1
```
alexsednz, line 5: Do we have this docker image released?

KarishmaBothara: yes.. I used it locally too

packages/types/src/interfaces/staking/definitions.ts
```diff
@@ -10,5 +10,6 @@ export default {
         Account: 'AccountId',
       },
     },
+    VecDeque: 'Vec<any>',
```
alexsednz, line 13: Do we have any other Vec type defined that it may collide with?

holygits (approved):
lgtm but at some point it would be good to get a test that `api.query.rewards.transactionFeePotHistory()` works since it's using the VecDeque type

packages/types/src/interfaces/staking/definitions.ts
KarishmaBothara: Check this https://github.com/cennznet/api-types/pull/14#discussion_r517693253
VecDeque should be generic so that can be used for all the types

KarishmaBothara:
> lgtm but at some point it would be good to get a test that `api.query.rewards.transactionFeePotHistory()` works since it's using the VecDeque type

I am looking at adding this in a test in this PR

packages/types/src/interfaces/staking/definitions.ts
KarishmaBothara: updated and remove it from here.

packages/types/src/interfaces/injects.ts
```diff
@@ -24,6 +25,7 @@ const _types = {
   other: {
     types: {
       ExtrinsicSignatureV4: CENNZnetExtrinsicSignatureV1,
+      VecDeque: VecAny,
```
KarishmaBothara, line 28: https://github.com/polkadot-js/api/blob/ffaea83e3eff25787fd1a1d0660d9d0bd7bdced6/packages/types/README.md

[[VecAny]] | This manages codec arrays, assuming that the inputs are already of type Codec
-- | --
https://github.com/polkadot-js/api/blob/33978d5007becbca5884862968eb778a866b6a94/packages/types/src/codec/VecAny.ts#L15

alexsednz (approved):
lgtm!

PR #209, opened 2020-11-17
support decoding of old block and current block on V37 chain

KarishmaBothara:
This PR is to address the issue decoding the blocks running on Azalea which supports old types (Extrinsic Signature). The fix was inspired by the way it was handled in https://github.com/polkadot-js/api/issues/2723
With typesBundle we can now let the API know which type definition to use for what version
I have tested it for some blocks on Azalea which had GA as well as Sylo extrinsic.. It decodes well.. Also tried to use send an extrinsic on v37 and tried to see if the block decodes well..

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,8 @@ jobs:
       - run:
           name: Run integration tests
           command: ./scripts/circleci/run_automation_test.sh -t integration
+          environment:
+          API_KEY: ${API_KEY}
   publish:
     docker:
       - image: circleci/node:10.16.3
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,8 +35,6 @@ jobs:
       - run:
           name: Run integration tests
           command: ./scripts/circleci/run_automation_test.sh -t integration
-          environment:
-          API_KEY: ${API_KEY}
   publish:
     docker:
       - image: circleci/node:10.16.3
--- a/packages/api/test/e2e/api.calls.e2e.ts
+++ b/packages/api/test/e2e/api.calls.e2e.ts
@@ -14,6 +14,10 @@
 import { Hash, Block, AccountId, EventRecord, AssetId } from '@cennznet/types'
 import { HeaderExtended } from '@polkadot/api-derive';
 import initApiPromise from '../../../../jest/initApiPromise';
+import {Api} from "@cennznet/api";
+import {SignedBlock} from "@polkadot/types/interfaces/runtime";
+import {Extrinsic} from "@polkadot/types/interfaces";
+import {Vec} from "@polkadot/types";
 
 describe('e2e api calls', () => {
   let api;
@@ -34,6 +38,25 @@ describe('e2e api calls', () => {
   it('Get correct block', async () => {
     const block: Block = await api.rpc.chain.getBlock(blockHash).then((r: any) => r.block);
     expect(block.header.hash.toString()).toEqual(blockHash.toString());
+  });
+
+  it('Query historical block from runtime version 36', async () => {
+    const blockNumber = 3759962; // old Azalea block at runtime version 36
+    const API_KEY = process.env.API_KEY;
+
+    const provider = `wss://node-6711773975684325376.jm.onfinality.io/ws?apikey=${API_KEY}`;
+    const apiV36 = await Api.create({provider});
+    const bHash = await apiV36.rpc.chain.getBlockHash(blockNumber);
+    const block: SignedBlock = await apiV36.rpc.chain.getBlock(bHash);
+    const extrinsicList: Vec<Extrinsic> = block.block.extrinsics;
+    expect(extrinsicList[0].method.section).toEqual('timestamp');
+    expect(extrinsicList[0].method.method).toEqual('set');
+    expect(extrinsicList[1].method.section).toEqual('finalityTracker');
+    expect(extrinsicList[1].method.method).toEqual('finalHint');
+    expect(extrinsicList[2].method.section).toEqual('syloE2Ee');
+    expect(extrinsicList[2].method.method).toEqual('replenishPkbs');
+    expect(extrinsicList[3].method.section).toEqual('syloE2Ee');
+    expect(extrinsicList[3].method.method).toEqual('withdrawPkbs');
   });
 
   it('Get correct validators', async () => {
--- a/packages/types/src/interfaces/extrinsic/v0/ExtrinsicPayloadV0.ts
+++ b/packages/types/src/interfaces/extrinsic/v0/ExtrinsicPayloadV0.ts
@@ -1,0 +1,145 @@
+// Copyright 2019-2020 Centrality Investments Limited & @polkadot/types authors & contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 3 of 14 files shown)


packages/types/src/interfaces/extrinsic/v0/ExtrinsicPayloadV0.ts
```diff
@@ -0,0 +1,146 @@
+// Copyright 2019 Centrality Investments Limited & @polkadot/types authors & contributors
```
KarishmaBothara, line 1: Re-Introduced the old Payload class we had when we supported v36

packages/types/src/interfaces/extrinsic/v0/ExtrinsicSignatureV0.ts
```diff
@@ -0,0 +1,279 @@
+// Copyright 2019 Centrality Investments Limited & @polkadot/types authors & contributors
```
KarishmaBothara, line 1: Re-Introduced the old Extrinsic Signature class we had when we supported v36

packages/types/src/interfaces/injects.ts
```diff
@@ -33,3 +34,25 @@ const _types = {
 
 // Unwind the nested type definitions into a flat map
 export default Object.values(_types).reduce((res, { types }): object => ({ ...res, ...types }), {});
+
+export const typesBundle: OverrideBundleType = {
```
KarishmaBothara, line 38: All the magic happens here

packages/types/src/interfaces/sylo/v0/index.ts
```diff
@@ -0,0 +1,129 @@
+// Copyright 2019 Centrality Investments Limited
```
KarishmaBothara, line 1: Re-Introduced the old Sylo types we had when we supported v36 to have SYLO extrinsics to decode fine

packages/types/src/interfaces/injects.ts
```diff
@@ -33,3 +34,25 @@ const _types = {
...
+      types: [
+        {
+          minmax: [0, 36],
+          types: {
+            ...syloTypes,
+            ExtrinsicSignatureV4: CENNZnetExtrinsicSignatureV0,
+          },
+        },
+        {
+          minmax: [37, 37],
```
alexsednz, line 50: Can we make this automatically include our future types > 37?

packages/types/src/interfaces/sylo/v0/index.ts
alexsednz: Which Sylo types have changed?

KarishmaBothara: I can see all the types(sylo) in the new API but they are auto generated from the definition and we don't have any test(Sylo).. I am investigating if the types we have currently are good for Sylo usage.. Trying to see how struct decodes...

packages/types/src/interfaces/injects.ts
KarishmaBothara: I will check if there is any option..

packages/types/src/interfaces/sylo/v0/index.ts
KarishmaBothara: I have checked polkadot-api/packages/types/src/interfaces, there is no way to set the last parameter in a struct.. i.e
constructor (registry: Registry, Types: S, value: V | Map<unknown, unknown> | unknown[] | string = {} as V, **jsonMap**: Map<keyof S, string> = new Map()) {

In Sylo types, for example

```
const GROUP_JSON_MAP = new Map([['groupId', 'group_id']]);

export class Group extends Struct {
  constructor(registry: Registry, value) {
    super(
      registry,
      {
        groupId: 'H256',
        members: Vec.with(Member),
        invites: Vec.with(PendingInvite),
        meta: Meta,
      },
      value,
      GROUP_JSON_MAP
    );
  }
}
```

there is no way to set json_map in definition..

packages/types/src/interfaces/injects.ts
KarishmaBothara: https://github.com/polkadot-js/api/blob/98b01a462dfc8d3280396d76c887ff8a3e1cd2e5/packages/types/src/types/registry.ts#L98
The minmax is of type [number, number] and has ? so can be undefined...
https://github.com/polkadot-js/api/blob/2c735fbc5397d482c331ff335c10faadb0adeadc/packages/types-known/src/spec/node-template.ts#L8

packages/api/test/e2e/api.calls.e2e.ts
```diff
@@ -36,6 +36,15 @@ describe('e2e api calls', () => {
     expect(block.header.hash.toString()).toEqual(blockHash.toString());
   });
 
+  it.skip('Query historical block', async () => {
```
holygits, line 39: this test could form a new set of e2e tests which are added each runtime upgrade / release to check backwards compatibility is maintained.

Is it possible to hard-code a v36 block to decode here for this test?

holygits (commented):
great PR!
Approach lgtm, please fix the test.

packages/api/test/e2e/api.calls.e2e.ts
KarishmaBothara: used onFinality node

packages/api/test/e2e/api.calls.e2e.ts
```diff
@@ -36,13 +40,21 @@ describe('e2e api calls', () => {
...
-  it.skip('Query historical block', async () => {
-    // const blockNumber = 2500; //local
-    const blockNumber = 3759962; //Azalea
-    const bHash = await api.rpc.chain.getBlockHash(blockNumber);
-    const block = await api.rpc.chain.getBlock(bHash);
-    console.log('Signed block::', block.toHuman());
-    console.log('Signed block::', block.toJSON());
+  it('Query historical block from runtime version 36', async () => {
+    const blockNumber = 3759962; // old Azalea block at runtime version 36
+    const provider = 'wss://node-6711773975684325376.jm.onfinality.io/ws?apikey=57eebed1-6b87-4fcf-8bb7-3d285e3eabff';
```
holygits, line 45: this is not good from a security standpoint (making the api key public). If there's no other way then I'm happy to merge this as it's an important fix

holygits (approved):

packages/api/test/e2e/api.calls.e2e.ts
KarishmaBothara: I see what you mean.. I will check if I can have one node running from runtime version 36 with docker, may be... I will investigate

holygits: Ideally we don't have to add a new docker image with each test, if we could keep the block data (e.g hex/json) in the test I think that would be more portable.
Otherwise some test where we upgrade a node's runtime, verify block, upgrade again, verify block, so on starting at v36 would be 🔥 and future proof

.circleci/config.yml
```diff
@@ -35,6 +35,8 @@ jobs:
       - run:
           name: Run integration tests
           command: ./scripts/circleci/run_automation_test.sh -t integration
+          environment:
+          API_KEY: ${API_KEY}
```
holygits, line 39: I don't think this change is required, it's just setting the variable to itself

KarishmaBothara: ok.. I'll remove this and let circle ci run.. thanks

PR #220, opened 2020-11-25
Fixed fee asset example while sending extrinsic

KarishmaBothara:
This PR fixes the docs example where transactionPayment option is used.. earlier the example worked, but the fees were not used from the fee asset (in this case CENNZ)..

### DIFF
--- a/docs/examples/promise/12_chose_fee_asset_to_pay_tx_fee/index.js
+++ b/docs/examples/promise/12_chose_fee_asset_to_pay_tx_fee/index.js
@@ -12,7 +12,7 @@ const AMOUNT = 10000;
 const FEE_ASSET_ID = 16000;
 const MIN_REQUIRED_POOL_BALANCE = 1000000;
 
-// Asset Id for CENNZ in Rimu
+// Asset Id for CENNZ in Nikau
 const CPAY = 16001;
 
 async function queryPoolBalance(api) {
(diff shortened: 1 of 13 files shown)


holygits (approved):
lgtm, is there a way to add this to test so we know it's working in future?

KarishmaBothara:
> lgtm, is there a way to add this to test so we know it's working in future?

It was working before.. but the fees getting deducted were CPAY.. Now I paid more attention on the events as I was expecting one event from cennzx as AssetBought

We also have a test for this in tx.e2e to send extrinsic with transactionPayment

docs/examples/promise/12_chose_fee_asset_to_pay_tx_fee/index.js
```diff
@@ -8,17 +8,12 @@ const { Keyring } = require('@polkadot/keyring');
...
-// some constants we are using in this sample
-const ALICE = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
-
-// BOB is pool liquidity provider
-const BOB = '5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty';
 const AMOUNT = 10000;
 const FEE_ASSET_ID = 16000;
 const MIN_REQUIRED_POOL_BALANCE = 1000000;
 
 // Asset Id for CENNZ in Rimu
```
alexsednz, line 15: Perhaps we need to change Rimu to testnet or Nikau

alexsednz (approved):

docs/examples/promise/12_chose_fee_asset_to_pay_tx_fee/index.js
KarishmaBothara: I am going thro' all the examples to fix the testnet to Nikau and check if any other comments needs to any modification

holygits (approved):

PR #222, opened 2020-11-26
Generating types from API repo

KarishmaBothara:
One place to generate types.. So that whenever we want to add new types, we don't have to change it in api-types and use it in api/types. The way it works is
cd types
yarn build
and then in auto generated files add // @ts-ignore  for RewardDestination, Staking and AssetOption as these types already exist in polkadot/types


holygits (approved):
cool! so we can archive api-types?

KarishmaBothara:
> cool! so we can archive api-types?

yes, I think so

PR #239, opened 2020-12-03
Fix dependencies in API package

KarishmaBothara:
@cennznet/util - is been used in https://github.com/cennznet/api.js/blob/674f3e22fa9dd4ff5a044d8127dc0c7628eeef1a/packages/api/src/derives/cennzx/utils/index.ts

Discovered this issue while using the new version in uncover project..
The reason for this issue is earlier @cennnznet/types had this dependencies.. But when we moved types generation in @cennznet/types.... we removed this dependency cause that was not required for @cennznet/types project.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -20,6 +20,7 @@
   },
   "dependencies": {
     "@cennznet/types": "^1.3.1",
+    "@cennznet/util": "^1.3.1",
     "@polkadot/api": "2.7.1",
     "@polkadot/metadata": "2.7.1",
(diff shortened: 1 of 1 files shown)


holygits (approved):
lgtm, would be good to figure out how this is not tested during our build step

alexsednz:
Can we also fix the warnings that Github Actions shows? Or possibly have a separate PR for that

KarishmaBothara:
> Can we also fix the warnings that Github Actions shows? Or possibly have a separate PR for that

may be in another PR..

PR #262, opened 2021-02-18
Prerelease/1.3.3 alpha.1

KarishmaBothara:
Corrected regex further...
The earlier one yielded - **version 1.3.3-alpha.1**
However we required 1.3.3-alpha.1
https://github.com/cennznet/api.js/runs/1923210924?check_suite_focus=true
Updated regex and tested locally
cat packages/api/package.json         | grep version         | head -1         | grep -E -o '[0-9|.]+\-*[a-z]*\.[0-9]+|]+'
tested with version **1.3.3** and **1.3.3-alpha.1**

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,7 +38,7 @@ jobs:
         echo ::set-output name=tag_name::$(cat packages/api/package.json \
         | grep version \
         | head -1 \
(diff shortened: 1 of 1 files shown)


holygits (approved):
if only there was an easier way to test these things 😅

PR #272, opened 2021-03-17
Copied staking derived queries from UI project

KarishmaBothara:
Earlier the derived queries were added in UI project... With this PR, the api will be decorated with proper derived queries which are used in UI. 
Few things to mention..

1. Updated eslint rule for duplicate import  - 
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md updated the duplicate import rule as we now have 
import type {xyz} from 'abc';
import {def} from 'abc'; 
false positive rule on import type - https://github.com/typescript-eslint/typescript-eslint/issues/2315

2. Updated prettier to latest version to support typescript 3.8 syntax (for example import type) 
https://github.com/prettier/prettier/issues/7263

3. For the new staking derived queries added, needed to add // @ts-ignore at few places as in parent (polkadot) same call is used with different argument... once we upgrade to latest polkadot version these can be removed...

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ It represents the closest reasonable ESLint configuration to this
 project's original TSLint configuration.
 
 We recommend eventually switching this configuration to extend from
-the recommended rulesets in typescript-eslint. 
+the recommended rulesets in typescript-eslint.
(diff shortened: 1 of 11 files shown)


holygits (approved):
lgtm

PR #276, opened 2021-04-12
Add NFT initial types/interfaces

holygits:


packages/types/src/interfaces/grandpa/definitions.ts
```diff
@@ -1,10 +0,0 @@
-export default {
-  rpc: {
-    restartVoter: {
```
KarishmaBothara, line 3: I think we need this rpc call? else we will get warning for grandpa.rpc.restartVoter does not exist?

holygits:
@KarishmaBothara I will split this PR into the clean up, update, and NFT module stuff

packages/types/src/interfaces/grandpa/definitions.ts
holygits: yeah, wasn't sure how to make it merge with the substrate definitions

packages/api/src/derives/staking/index.ts
```diff
@@ -1,6 +1,5 @@
 export { queryStakingAccountInfo as accountInfo } from './stakingAccount';
 export * from './overview';
-export * from './query';
```
holygits, line 3: doesn't build dues to type issues, it isn't used from what I can see so opting to remove

tsconfig.json
```diff
@@ -22,7 +22,9 @@
       "@cennznet/types/*": ["packages/types/src/*"],
       "@cennznet/types": ["packages/types/src/index"],
       "@cennznet/util/*": ["packages/util/src/*"],
-      "@cennznet/util": ["packages/util/src/index"]
+      "@cennznet/util": ["packages/util/src/index"],
+      "@polkadot/api/augment": ["packages/types/src/interfaces/augment-api.ts"],
```
holygits, line 26: fixes the `//@ts-ignores` in augment-types

packages/api/src/derives/staking/index.ts
KarishmaBothara: added this so that we can remove decorating api with staking derived queries in apps/UI project..

KarishmaBothara (approved):

packages/api/src/derives/staking/index.ts
holygits: ah so will this break the UI?

KarishmaBothara: no.. it won't... the UI is decorating it atm.

PR #290, opened 2021-05-06
Update release build

holygits:
Tidy up api README

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,7 +1,6 @@
 module.exports = {
   extends: '@polkadot/dev/config/babel-config-cjs.cjs',
   "plugins": [
-    "@babel/plugin-proposal-numeric-separator",
     ["@babel/plugin-proposal-private-methods", { "loose": true }]
   ]
 };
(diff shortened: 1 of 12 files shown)


babel.config.js
```diff
@@ -1,7 +1,6 @@
 module.exports = {
   extends: '@polkadot/dev/config/babel-config-cjs.cjs',
   "plugins": [
-    "@babel/plugin-proposal-numeric-separator",
```
holygits, line 4: included in `'@polkadot/dev/config/babel-config-cjs.cjs',`

KarishmaBothara (approved):

packages/api/src/Api.ts
```diff
@@ -25,7 +25,7 @@ import { getProvider } from './util/getProvider';
 import { getTimeout } from './util/getTimeout';
 
 export class Api extends ApiPromise {
-  static create(options: ApiOptions = {}): Promise<Api> {
+  static async create(options: ApiOptions = {}): Promise<Api> {
```
KarishmaBothara: not sure if this has to be async..
async and await are used together and 
when a function returns a new Promise it need not be async..

holygits: there were 2 conflicting lints. one requires `async` to use an `await` inside, the other requires `async` when function returns a promise

Issue #244, opened 2020-12-07
improve tx payment interface


holygits:
multi-currency fee payment API could provide a better interface to take % slip vs. guessing a large enough amount, quite clunky atm.

current API:
https://github.com/cennznet/api.js/blob/b803ce8169994751a43492a344fab2d1a7ddbbb4/packages/api/test/e2e/tx.e2e.ts#L160-L163

nicer API:
```ts
api.tx.genericAsset
  .transfer(spendingAssetId, bob.address, 100)
  .paymentOpts({ feeAssetId: CENNZ, slippage: 3 });
```

KarishmaBothara:
@eupston -  check this commit - https://github.com/cennznet/api.js/commit/4dd7768b77ee6cfaf958122d604de6186b8ab8a4
It will give you idea of how to decorate the extrinsic to support paymentOpts... here you can see fee is decorated on extrinsic...            

            ```
const tx = api.tx.genericAsset.create({
                initialIssuance: 100,
            });
            const fee = await tx.fee(sender.address);
```

**Comment by eupston** [comment]
amazing thank you @KarishmaBothara. It looks like the custom cennznet extrinsics decorators were removed when updating to polkadot v1.1.1, although yeah this definitely gives me a good idea of how to reimplement.

**Comment by eupston** [comment]
@holygits just thinking about the design of this a bit more and currently we rely on creating the `transactionPayment` when going to `signAndSend`. So was your idea to wrap `signAndSend` in `paymentOpts`? Definitely doable but will require passing down a few extra params such as `assetOwner` and `tip`. 

Also would slippage percentage be based on the estimated fee of the provided `feeAssetId`?

**Comment by holygits** [comment]
So there's a couple ideas here.
Some might be infeasible and bigger than just this interface.

1) `api` could get the current price of the transaction in CPAY and user passes a slippage `%` on top, knowing the price estimate is accurate they just say they don't mind a maximum price shift of x%
internally I imagine this would query the current tx price and calculate current price * (1 +slippage/100)

To use this interface in it's current state the developer needs to query the price anyway so we could make it simpler for them by internalizing that...

2) It would be nice if the method chaining continued when setting other options
rather than passing to `signAndSend`

```js
api.tx.module.method
  .nonce(11)
  .paymentOpts({ assetId, slippage })
  .signAndSend(//...)
```

PR #299, opened 2021-05-07
Prerelease/2.0.0 alpha.2

holygits:

PR #316, opened 2021-05-25
improved extrinsic payment options interface (#244)

eupston:
@holygits here's the initial improvement to the interface as discuss we may want to pass the instance of the API object down to the extrinsic in another PR which would give us quite a lot of flexibility and potential for more abstraction.

### DIFF
--- a/packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
+++ b/packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
@@ -1,0 +1,89 @@
+import { GenericExtrinsic } from '@polkadot/types';
+import { SignatureOptions } from '@cennznet/types/interfaces/extrinsic/types';
+import { AnyNumber, AnyU8a, ExtrinsicPayloadValue, IKeyringPair, Registry } from '@polkadot/types/types';
+import { Address, Call } from '@polkadot/types/interfaces/runtime';
+import { ExtrinsicValueV4 } from '@polkadot/types/extrinsic/v4/Extrinsic';
+import { Api } from '@cennznet/api';
+import { PaymentOptions } from '@cennznet/api/derives/types';
+
+export default class CENNZnetExtrinsic extends GenericExtrinsic{
+  private signaturePayloadOptions: SignatureOptions | ExtrinsicPayloadValue;
+
+  constructor(registry: Registry, value?: GenericExtrinsic | ExtrinsicValueV4 | AnyU8a | Call) {
+    super(registry, value);
+    this.signaturePayloadOptions = {
+      blockHash: undefined,
+      era: undefined,
+      genesisHash: undefined,
+      method: undefined,
+      specVersion: undefined,
+      nonce: undefined,
+      runtimeVersion: undefined,
+      signedExtensions: undefined,
+      signer: undefined,
+      tip: undefined,
+      transactionPayment: undefined,
+      transactionVersion: undefined
+    };
+  }
+
+  setNonce(value: AnyNumber){
+    this.signaturePayloadOptions.nonce = value;
+    return this;
+  }
+
+  addSignature(signer: Address | Uint8Array | string, signature: Uint8Array | string, payload: ExtrinsicPayloadValue | Uint8Array | string): GenericExtrinsic {
+    const mergeDefinedObjects = (A, B) => {
+      const res = {};
+      Object.keys({...A,...B}).map(key => {
+        res[key] = A[key] || B[key];
+      });
+      return res as ExtrinsicPayloadValue;
+    };
+
+    const mergedPayload = mergeDefinedObjects(payload, this.signaturePayloadOptions);
+    return super.addSignature(signer, signature, mergedPayload);
+  }
+
+  /**
+   * @description Sign the extrinsic with a specific keypair
+   */
+  sign(account: IKeyringPair, options: SignatureOptions): GenericExtrinsic {
+    const mergeDefinedObjects = (A, B) => {
+      const res = {};
+      Object.keys({...A,...B}).map(key => {
+          res[key] = A[key] || B[key];
+      });
+      return res as SignatureOptions;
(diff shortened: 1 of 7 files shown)


packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
```diff
@@ -0,0 +1,110 @@
...
+        return '';
+      }, toU8a(isBare?: BareOpts): Uint8Array {
+        return undefined;
+      },
+        extrinsic : this,
+        userFeeAssetId: feeAssetId,
+        maxPayment: _maxPayment
+    }
+    const getEstimatedFee = async () => {
+      const estimatedFee = await api.derive.fees.estimateFee(estimateFeeParams)
```
holygits, line 90: so if I understand correctly `EstimateFeeParams` doesn't impl Codec so we implemented it here? I think we can relax that requirement in the types and already did so on the `master` branch cc @KarishmaBothara 
should make this more readable

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
```diff
@@ -0,0 +1,110 @@
...
+      return res as SignatureOptions;
+    };
+    const mergedSignatureOpts = mergeDefinedObjects(options, this.signaturePayloadOptions);
+    return super.sign(account, mergedSignatureOpts);
+  }
+
+  /**
+   * @description Adds payment options such as slippage and tip to the Extrinsic signature
+   */
+  paymentOpts(api: Api, feeAssetId: number | string, slippage = 0, tip = 0): this{
```
holygits, line 67: let's call it `setPaymentOpts` for consistency
also providing de-structuring will help for readability

```suggestion
  setPaymentOpts(api: Api, { feeAssetId: number | string, slippage = 0, tip = 0 }): this{
```

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
```diff
@@ -0,0 +1,110 @@
...
+      const estimatedFee = await api.derive.fees.estimateFee(estimateFeeParams)
+      return estimatedFee;
+    }
+    getEstimatedFee()
+      .then((estimatedfee) => {
+      const maxPayment = parseFloat(estimatedfee.toString()) * (1 + slippage)
+      const assetId = this.registry.createType('AssetId', feeAssetId);
+      const feeExchange = this.registry.createType('FeeExchange', { assetId, maxPayment }, 0);
+      const transactionPayment = this.registry.createType('ChargeTransactionPayment', {tip: tip, feeExchange});
+      if ('transactionPayment' in this.signaturePayloadOptions) {
```
holygits, line 99: wondering when is this not `true` (genuine question)?

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
```diff
@@ -0,0 +1,110 @@
...
+        userFeeAssetId: feeAssetId,
+        maxPayment: _maxPayment
+    }
+    const getEstimatedFee = async () => {
+      const estimatedFee = await api.derive.fees.estimateFee(estimateFeeParams)
+      return estimatedFee;
+    }
+    getEstimatedFee()
+      .then((estimatedfee) => {
+      const maxPayment = parseFloat(estimatedfee.toString()) * (1 + slippage)
```
holygits, line 95:
```suggestion
      const maxPayment = parseFloat(estimatedfee.toString()) * (1 + slippage);
```

holygits (commented):
nice. general direction lgtm. thinking we just need a bunch of tests to check the compatibility with the current `signAndSend` implementation e.g.

`tx.setNonce(nonce).signAndSend(nonce)` <- what should be our default behaviour

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
eupston: yup sure can change

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
eupston: yeah definitely so `EstimateFeeParams` shouldn't need to extend `Codec` then? looks like it was added for Prerelease/2.0.0 alpha.2 (#299)

holygits: yeah for `derive` queries arguments don't need to have `Codec` so we decided to relax that typing

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
eupston: oo just when I'm using one object to store both `SignatureOptions | ExtrinsicPayloadValue` and `ExtrinsicPayloadValue` doesn't contain transactionPayment, but thinking about it now might be cleaner just to split out the signature and payload into two objects.

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
eupston: okay sweet i'll change that

eupston:
> nice. general direction lgtm. thinking we just need a bunch of tests to check the compatibility with the current `signAndSend` implementation e.g.
> 
> `tx.setNonce(nonce).signAndSend(nonce)` <- what should be our default behaviour

yup for sure I added two test variations so far, one with the nonce being passed into `signAndSign` directly and one setting it via chaining. I'd say chaining should be default? means less overhead for user. In scenario above, if user sets nonce via chain and then passes in the nonce again to signAndSend it will use the one passed in. You reckon we should have a test to ensure it does use the latter?

eupston:
closing PR and reopening changes on master as integration tests on develop aren't compatible with the latest cennznet version 1.5.0.

PR #318, opened 2021-05-28
Improve tx payment_interface(#244)

eupston:
continuation of the PR to develop branch:
https://github.com/cennznet/api.js/pull/316

### DIFF
--- a/packages/api/src/derives/fees/estimateFee.ts
+++ b/packages/api/src/derives/fees/estimateFee.ts
@@ -36,7 +36,10 @@ import { catchError, first, map, switchMap } from 'rxjs/operators';
 //    const maxPayment = 'xxx' // Max amount user is ok to pay in 'fee asset'
 //    const feeFromQuery = await api.derive.fees.estimateFee({extrinsic, userFeeAssetId, maxPayment}) // this will only be successful if their is enough liquidity of users asset in exchange pool
 
-export function estimateFee(instanceId: string, api: ApiInterfaceRx) {
+export function estimateFee(
+  instanceId: string,
+  api: ApiInterfaceRx
+): ({ extrinsic, userFeeAssetId, maxPayment }: EstimateFeeParams) => Observable<Balance | Error> {
   // We generate fake signature data here to ensure the estimated fee will correctly match the fee paid when the extrinsic is signed by a user.
   // This is because fees are currently based on the byte length of the extrinsic
   return ({ extrinsic, userFeeAssetId, maxPayment }: EstimateFeeParams): Observable<Balance | Error> => {
--- a/packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
+++ b/packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
@@ -1,0 +1,82 @@
+import { GenericExtrinsic, UInt } from '@polkadot/types';
+import { SignatureOptions } from '@cennznet/types/interfaces/extrinsic/types';
+import { AnyU8a, ExtrinsicPayloadValue, IKeyringPair, Registry } from '@polkadot/types/types';
+import { Address, Call } from '@polkadot/types/interfaces/runtime';
+import { ExtrinsicValueV4 } from '@polkadot/types/extrinsic/v4/Extrinsic';
+import { Api } from '@cennznet/api';
+import { PaymentOptions } from '@cennznet/api/derives/types';
+import { Constructor } from '@polkadot/util/types';
+
+export default class CENNZnetExtrinsic extends GenericExtrinsic{
+  private signaturePayloadOptions: SignatureOptions | ExtrinsicPayloadValue;
+
+  constructor(registry: Registry, value?: GenericExtrinsic | ExtrinsicValueV4 | AnyU8a | Call) {
+    super(registry, value);
+    this.signaturePayloadOptions = {
+      blockHash: undefined,
+      genesisHash: undefined,
+      method: undefined,
+      specVersion: undefined,
+      nonce: undefined,
+      runtimeVersion: undefined,
+      transactionPayment: undefined,
+      transactionVersion: undefined
+    };
+  }
+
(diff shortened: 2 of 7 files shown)


packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
```diff
@@ -0,0 +1,89 @@
...
+   */
+  sign(account: IKeyringPair, options: SignatureOptions): GenericExtrinsic {
+    const mergeDefinedObjects = (A, B) => {
+      const res = {};
+      Object.keys({...A,...B}).map(key => {
+          res[key] = A[key] || B[key];
+      });
+      return res as SignatureOptions;
+    };
+    const mergedSignatureOpts = mergeDefinedObjects(options, this.signaturePayloadOptions);
```
KarishmaBothara, line 59: wondering if we can use it as
const mergedSignatureOpts = { ...options, ...this.signaturePayloadOptions };?

packages/types/src/interfaces/injects.ts
```diff
@@ -52,6 +53,9 @@ export const typesBundle: OverrideBundleType = {
         {
           minmax: [37, undefined],
           types: {
+            // In our staking runtime module, we use StakingLedgerTo223 definition of polkadot's staking ledger
+            StakingLedger: 'StakingLedgerTo223',
```
KarishmaBothara, line 57: in this version of master may be we don't have StakingLedgerTo223
as master is still using 2.7.1

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
eupston: na unfortunately not because we want to only merge the props that are not undefined in `signaturePayloadOptions`

packages/types/src/interfaces/injects.ts
eupston: ahh true think this snuck over from develop branch when I cherry picked, will remove

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
```diff
@@ -0,0 +1,89 @@
...
+      era: undefined,
+      genesisHash: undefined,
+      method: undefined,
+      specVersion: undefined,
+      nonce: undefined,
+      runtimeVersion: undefined,
+      signedExtensions: undefined,
+      signer: undefined,
+      tip: undefined,
+      transactionPayment: undefined,
```
KarishmaBothara, line 25: export interface SignatureOptions {
    blockHash: Uint8Array | string;
    era?: IExtrinsicEra;
    genesisHash: Uint8Array | string;
    nonce: AnyNumber;
    runtimeVersion: IRuntimeVersion;
    signedExtensions?: string[];
    signer?: Signer;
    tip?: AnyNumber;
}

era, signedExtension, signer, tip and transactionPayment are optional field and its ok even if we don't explicit set them to undefined

KarishmaBothara:
nice work @eupston 👍

packages/api/src/derives/fees/estimateFee.ts
```diff
@@ -36,7 +36,7 @@ import { catchError, first, map, switchMap } from 'rxjs/operators';
 //    const maxPayment = 'xxx' // Max amount user is ok to pay in 'fee asset'
 //    const feeFromQuery = await api.derive.fees.estimateFee({extrinsic, userFeeAssetId, maxPayment}) // this will only be successful if their is enough liquidity of users asset in exchange pool
 
-export function estimateFee(instanceId: string, api: ApiInterfaceRx) {
+export function estimateFee(instanceId: string, api: ApiInterfaceRx): any {
```
KarishmaBothara, line 39: can we try some ways to specify type for this... 
Observable<?> or Observable<any>

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
eupston: yup sweet can update that

packages/api/src/derives/fees/estimateFee.ts
eupston: yup seems possible just need to change to a function signature that returns Observable

holygits (approved):
lgvtm. Could you add one test to document the behaviour
of

```js
setNonce(nonce1).signAndSend(nonce2)
```

KarishmaBothara (approved):

eupston:
@holygits So done a bit of a deep dive on how nonce is set, and found we won't be able to override the nonce from the `GenericExtrinsic` class so I've removed the ability to set it from the interface for the time being.

The nonce is set in the `SubmittableExtrinisic` which by the time our custom `CENNZnetExtrinsic` goes to modify the nonce it has already been used and signed, because `SubmittableExtrinisic` is at the top layer of the inheritance chain. So not possible to change with current imp but at least means we know the nonce is working properly lol. 

Another interesting note about the nonce is that if no nonce is passed in it will actually create one automatically with the address used for the transaction. see signingInfo [here](https://github.com/polkadot-js/api/blob/9b9581487d65351bd568df050011ed580a596071/packages/api-derive/src/tx/signingInfo.ts#L62).
 So really you only need to set the nonce if you want to sign it with an another account that isn't used for the transaction.  
 
I believe if we extend `SubmittableExtrinisic` as well as the `createClass` function with our own custom `CENNZnetSubmittableExtrinisic` we may be able to inject the nonce via chaining. However it is quite tricky because `SubmittableExtrinisic` is never declared as a class, and all implementation happens in `createClass`. So not quite sure how we'd override `signAndSend` and inject our custom nonce and then calling super to hook into the `createClass` implementation of `SubmittableExtrinisic`. May also be able to extend `signingInfo`....

So anyway think it may be best left to another PR, which at the same time could possible look into attaching API instance to the extrinsic. :)

holygits:
@eupston that makes sense! in that case keen to merge this

PR #326, opened 2021-06-03
Nft open listings (#320)

### LINKED ISSUE #320: NFT open listings
Make a derive to list all tokens for sale (listing ids) in a collection

```js
const listings = await api.query.nft.openCollectionListings.entries(collectionId);

listings.forEach(([key, listing]) => {
  console.log('key arguments:', key.args.map((k) => k.toHuman()));
  console.log('     listing:', listing.toHuman());
});
```

eupston:
Derived query for retrieving all tokens that have an open listings with the given collection ID. closes #320

### DIFF
--- a/packages/api/src/derives/nft/openCollectionListings.ts
+++ b/packages/api/src/derives/nft/openCollectionListings.ts
@@ -1,0 +1,88 @@
+// Copyright 2019-2021 Centrality Investments Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { Observable, combineLatest } from 'rxjs';
+import { map, switchMap, reduce, mergeAll, first } from 'rxjs/operators';
+import { ApiInterfaceRx } from '@cennznet/api/types';
+import { DeriveTokenInfo } from '@cennznet/api/derives/nft/types';
+import { Listing, TokenId } from '@cennznet/types';
+import { Option } from '@polkadot/types/codec/Option';
+import { UInt } from '@polkadot/types';
+
+/**
+ * Gets all tokens in a collection that have an open listing
+ *
+ * @param collectionId  The collection Id value
+ *
+ * @returns List of tokens that have open listings in given collection
+ */
+export function openCollectionListings(instanceId: string, api: ApiInterfaceRx) {
+  return (collectionId: string): Observable<DeriveTokenInfo[]> => {
+    return api.query.nft.openCollectionListings.keys(collectionId).pipe(
+      switchMap(
+        (storageKeys): Observable<DeriveTokenInfo[]> => {
+          const listingIDs = storageKeys.map((storageKey) => {
+            return storageKey.args.map((k) => k.toHuman())[1];
+          });
+          return api.query.nft.listings.multi(listingIDs.map((listingID) => listingID)).pipe(
+            switchMap(
+              (listingsCodec: Option<Listing>[]): Observable<DeriveTokenInfo[]> => {
+                const tokenIds = listingsCodec.map((listingCodec) => {
+                  const listing: Listing = listingCodec.unwrapOrDefault();
+                  const tokenID: TokenId = listing.isAuction
+                    ? listing.asAuction.toJSON().tokens[0]
+                    : listing.asFixedPrice.toJSON().tokens[0];
+                  return tokenID;
+                });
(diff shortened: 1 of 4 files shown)


packages/api/src/derives/nft/openCollectionListings.ts
```diff
@@ -0,0 +1,65 @@
...
+        (storageKeys): Observable<any> => {
+          const listingIDs = storageKeys.map((storageKey) => {
+            return storageKey.args.map((k) => k.toHuman())[1];
+          });
+          return api.query.nft.listings.multi(listingIDs.map((listingID) => listingID)).pipe(
+            switchMap(
+              (listingsCodec: Option<Listing>[]): Observable<any> => {
+                const tokenIds = listingsCodec.map((listingCodec) => {
+                  const listing: Listing = listingCodec.unwrapOrDefault();
+                  const tokenID: TokenId = listing.asAuction.toJSON().tokens[0];
```
holygits, line 42: listings can happen for `fixedPriceSale`s also

holygits (commented):
thinking we should return the listing IDs along with the token ID otherwise user will need to write similar code to figure it out

packages/api/src/derives/nft/openCollectionListings.ts
eupston: ahh right cool, i'll add that as well

eupston:
> thinking we should return the listing IDs along with the token ID otherwise user will need to write similar code to figure it out

ahh yup for sure had that in back of my mind, maybe a `listing ID`:`token Info` map then

packages/api/src/derives/nft/openCollectionListings.ts
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @param collectionId  The collection Id value
+ *
+ * @returns List of tokens that have open listings in given collection
+ */
+export function openCollectionListings(instanceId: string, api: ApiInterfaceRx) {
+  return (collectionId: string): Observable<DeriveTokenInfo[]> => {
+    return api.query.nft.openCollectionListings.keys(collectionId).pipe(
+      switchMap(
+        (storageKeys): Observable<any> => {
```
KarishmaBothara, line 33: Observable<DeriveTokenInfo[]>?

packages/api/test/e2e/nft.e2e.ts
```diff
@@ -64,6 +66,41 @@ afterAll(async () => {
 describe('NFTs', () => {
   let collectionId: number, collectionId2: number;
 
+  beforeEach(async done => {
```
KarishmaBothara, line 69: do we want create a collection and mint series - beforeAll (one time) or beforeEach...
BeforeEach will end up creating collection before each test?

packages/api/src/derives/nft/openCollectionListings.ts
eupston: oo yup i'll update that

packages/api/test/e2e/nft.e2e.ts
eupston: Yeah I guess my thinking is that each test should really have their own collection and series etc incase they modify it during the test then it won't effect other tests. However right now we have a lot of interdependency between tests which isn't great. 

Thought it might be worth leaving in `BeforeEach` even though only one test really utilises it so we have example of how to write subsequent tests and possibly refactor the current in the future.

eupston:
all issues should be resolved and closes issue #320

KarishmaBothara (approved):
lgtm

packages/api/test/e2e/nft.e2e.ts
```diff
@@ -247,41 +285,31 @@ describe('NFTs', () => {
...
-      {
-        collectionId: 1,
-        seriesId: 0,
-        serialNumber: 1,
-      },
-      {
-        collectionId: 1,
-        seriesId: 0,
-        serialNumber: 2,
+        "collectionId": 1,
```
holygits, line 299: nit: wondering if we need the `"` here?

packages/api/test/e2e/nft.e2e.ts
```diff
@@ -433,4 +451,63 @@ describe('NFTs', () => {
...
+                    seriesId: 0,
+                    serialNumber: 0,
+                  },
+                  {
+                    collectionId: globalCollectionId,
+                    seriesId: 0,
+                    serialNumber: 1,
+                  }
+                ]);
+                //Ensure Listing Id correctly matches token Id
```
holygits, line 495:
```suggestion
                // Ensure Listing Id correctly matches token Id
```
space after `//` (we should have a lint for this)

packages/api/test/e2e/nft.e2e.ts
```diff
@@ -360,17 +378,17 @@ describe('NFTs', () => {
     let ownedTokens = (await api.rpc.nft.collectedTokens(collectionId, tokenOwner.address));
     expect(ownedTokens.toJSON()).toEqual([
       {
-        collectionId: 0,
+        collectionId: collectionId,
```
holygits, line 381:
```suggestion
        collectionId,
```

holygits (approved):

packages/api/test/e2e/nft.e2e.ts
eupston: sweet yup can remove

PR #327, opened 2021-06-08
added example to use api with combineLatest to query different attributes

KarishmaBothara:
Good example to have it in repo for anyone who wants keep on subscription of few query when new block is added


holygits (approved):

eupston (approved):
nice lgtm

Issue #310, opened 2021-05-20
Don't break with new types


holygits:
When runtime gets a new module or types API refuses to connect because it's missing those types.
The recent 1.5.0/v40 release added new types in a backwards compatible way for the protocol (semver minor version bump)

this project should provide the same backward compatibility guarantee
e.g. it would be better if it failed to encode/decode new types rather than outright refuse to connect.

Overall this creates a bad experience for dapps which must update to handle it.

here's an example after recent 1.5.0/v40 upgrade

<img width="895" alt="Screen Shot 2021-05-25 at 15 31 32" src="https://user-images.githubusercontent.com/5133901/119435755-e7ec3000-bd6e-11eb-8c99-dd8f093cf9e8.png">

holygits:
DApps could register for the endpoints they use e.g. `Api.connect(['system', 'cennzx', 'nft'])` might even help with the metadata decoding issue

KarishmaBothara:
> DApps could register for the apis they use e.g. `Api.connect(['system', 'cennzx', 'nft'])` might even help with the metadata decoding issue

this is a good idea... if the list is empty, it will try to use the whole metadata for that genesisHash and specVersion... else use filtered metadata..

PR #331, opened 2021-06-16
First step to let user use custom metadata without specifying genesis…

KarishmaBothara:
Closes https://github.com/cennznet/api.js/issues/310
Options supplied to create API instance, can now include two more option..

1. fullMeta: boolean - if false (use minimum metadata)
2. modules: string[] - use metadata only for these runtime module, user is interested in eg.. [System, GenericAsset,...]

Things to do.
~~Add the same logic for Rx API and test it
Add unit test to getMetadata function
Check extrinsic works fine~~

~~This is still wip PR~~


packages/api/src/util/getMetadata.ts
```diff
@@ -0,0 +1,44 @@
...
+    providerInterface = provider || new WsProvider();
+  }
+  const registry = new TypeRegistry();
+  const rpcCore = new RpcCore('1', registry, providerInterface);
+  const meta: Metadata = await rpcCore.state.getMetadata().toPromise();
+  const magicNumber = meta.magicNumber;
+  const modules = meta.asLatest.modules.toArray();
+
+  const extrinsic = meta.asLatest.extrinsic;
+  const filteredModule = modules.filter((mod) => keepMetaFor.find((a) => a === mod.name.toString()));
```
holygits, line 28: we should do something nicer here like provide constants or normalise the string so that there's no confusion for users if they write 'system' or 'System' or 'System '

holygits (commented):
We'd want some tests on getMetadata to check what happens if you give a bad module name etc.
also a test to show this can solve the error in #310 would be great

The ideal outcome from this is that a user can do
```js
api.connect({ provider: ..., modules: ['GenericAsset'] })
```
to only interact with generic assets
I think in this case we should make some modules mandatory like 'System' and 'TransactionPayment' are always provided

packages/api/src/util/getMetadata.ts
KarishmaBothara: added this change

packages/api/src/util/getMetadata.ts
```diff
@@ -0,0 +1,44 @@
...
+  keepMetaFor: string[] = KEEP
+): Promise<Record<string, string>> {
+  let providerInterface: ProviderInterface;
+  if (typeof provider === 'string') {
+    providerInterface = getProvider(provider);
+  } else {
+    providerInterface = provider || new WsProvider();
+  }
+  const registry = new TypeRegistry();
+  const rpcCore = new RpcCore('1', registry, providerInterface);
```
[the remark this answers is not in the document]
KarishmaBothara: RpcCore accepts the first argument as instanceId..
https://github.com/polkadot-js/api/blob/master/packages/rpc-core/src/index.ts#L102
from API... whenever a new instance of API is created, this instance id is incremented as in https://github.com/polkadot-js/api/blob/master/packages/api/src/base/Decorate.ts#L136 and is feed to rpcCore object...

KarishmaBothara:
> We'd want some tests on getMetadata to check what happens if you give a bad module name etc.
> also a test to show this can solve the error in #310 would be great
> 
> The ideal outcome from this is that a user can do
> 
> ```js
> api.connect({ provider: ..., modules: ['GenericAsset'] })
> ```
> 
> to only interact with generic assets
> I think in this case we should make some modules mandatory like 'System' and 'TransactionPayment' are always provided

This can solve https://github.com/cennznet/api.js/issues/310 would mean.. I have a runtime upgrade with new types and I don't use that module while creating an API.... cause the current API will always contain all the types for the current runtime..

packages/api/test/e2e/api.create.e2e.ts
```diff
@@ -102,6 +114,41 @@ describe('e2e api create', () => {
     expect(extPayload.toHuman().transactionPayment).toEqual({tip: '2.0000 pUnit', feeExchange:{ FeeExchangeV1: { assetId: '16,000', maxPayment: '50.0000 mUnit' }}});
   });
 
+  it ('Should create api instance with custom metadata and execute transfer extrinsic', async done => {
+    const provider = config.wsProvider[`${process.env.TEST_TYPE}`];
+    api = await Api.create({provider, modules:['System', 'Timestamp ', ' TransactionPayment', 'GenericAsset', 'UnKnown']});
```
KarishmaBothara, line 119: Unknown Module would just get filtered..won't break anything

packages/api/test/e2e/rxapi.create.e2e.ts
```diff
@@ -32,6 +34,26 @@ describe('e2e rx api create', () => {
...
+    const stakingAssetId = await api.query.genericAsset.stakingAssetId();
+    expect(api.query.nft).toBeUndefined();
+    expect(stakingAssetId.toNumber()).toBeGreaterThan(0);
+    await api.disconnect();
+    done();
+  });
+
+  it ('Should create rx api instance with custom metadata', async done => {
+    const provider = config.wsProvider[`${process.env.TEST_TYPE}`];
+    const api = await Api.create({provider, modules:['TransactionPayment', 'GenericAsset']});
```
KarishmaBothara, line 49: when user specifies only module they are interested in... the mandatory module like System, TimeStamp would be auto included

eupston (approved):
lgtm. Only question is if cennzx module gets updated with a new type lets say, does that mean the whole cennzx module will be out of action for the user until we update the static metadata? 

Better than not being able to connect to the whole API of course.

KarishmaBothara:
> lgtm. Only question is if cennzx module gets updated with a new type lets say, does that mean the whole cennzx module will be out of action for the user until we update the static metadata?
> 
> Better than not being able to connect to the whole API of course.

If you don't include it as a part of metadata then it should work fine... But when you say 
api = await Api.create({provider, modules:['Cennzx']})
This can break... This PR will help when you introduce a runtime module in node... DApp user can still use the old API with upgraded node if they choose to use fullMeta = false and specify the custom modules they want to use...

packages/api/src/types.ts
```diff
@@ -32,6 +32,10 @@ export interface ApiOptions extends Pick<ApiOptionsBase, Exclude<keyof ApiOption
    * default 10000 ms, 0 indicates no limit
    */
   timeout?: number;
+
```
holygits, line 35: can you please add some doc comments here so developers get the nice hints when using this package

packages/api/src/util/getMetadata.ts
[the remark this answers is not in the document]
holygits: ok add a little comment so it's not 'magic' number

packages/api/src/util/getMetadata.ts
```diff
@@ -0,0 +1,50 @@
...
+import { TypeRegistry } from '@polkadot/types/create';
+import { RpcCore } from '@polkadot/rpc-core';
+import { MetadataVersioned } from '@polkadot/metadata/MetadataVersioned';
+import { Metadata } from '@polkadot/metadata';
+import { ProviderInterface } from '@polkadot/rpc-provider/types';
+import { WsProvider } from '@polkadot/rpc-provider';
+
+const KEEP = ['genericasset', 'cennzx'];
+const MUST = ['system', 'timestamp', 'transactionpayment'];
+export async function getMetadata(
```
holygits, line 11: can you doc comment this please

what's the difference between keep and must?

I think this can create some issue if user wants to add a module `modules: ['staking']` it will break functionality for genericasset + cennzx which worked before implicitly.

packages/api/src/util/getMetadata.ts
```diff
@@ -0,0 +1,50 @@
...
+  provider: string | ProviderInterface,
+  keepModules: string[] = KEEP
+): Promise<Record<string, string>> {
+  const useModules = keepModules.map(module => module.trim().toLowerCase());
+  MUST.map(mandatoryModule => {
+    if (!useModules.includes(mandatoryModule)) {
+      useModules.push(mandatoryModule);
+    }
```
holygits, line 19: does it make sense to use a set if we want unique members?

packages/api/src/util/getMetadata.ts
```diff
@@ -0,0 +1,50 @@
...
+  const extrinsic = meta.asLatest.extrinsic;
+  const filteredModule = modules.filter((mod) => useModules.find((a) => a === mod.name.toString().toLowerCase()));
+  const filteredModuleMetadataLatest = registry.createType('Vec<ModuleMetadataLatest>', filteredModule);
+  const metadataSlim = registry.createType('MetadataLatest', {
+    extrinsic,
+    modules: filteredModuleMetadataLatest,
+  });
+  const mVersionedSlim = new MetadataVersioned(registry, {
+    magicNumber: magicNumber,
+    metadata: registry.createType('MetadataAll', metadataSlim, 12),
```
holygits, line 42: why `12`? does this need to change? if so when?

packages/api/test/e2e/getMetadata.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {getMetadata} from "../../src/util/getMetadata";
+import {TypeRegistry} from "@polkadot/types/create";
+import {MetadataVersioned} from "@polkadot/metadata/MetadataVersioned";
+
+describe.skip('getMetadata()',  () => {
```
holygits, line 19: did you mean to unskip this?

KarishmaBothara: yes..  I am trying to run this as a part of e2e test... looking at this currently

packages/api/src/util/getMetadata.ts
KarishmaBothara: no.. m going to change this so that we don't need to touch this piece

PR #338, opened 2021-07-01
updated with useCennznet to connect to extension and return all accounts and api instance

### LINKED ISSUE #324: React hook support + extension integration
Provide a one-liner for dapp developers to connect with extension get available accounts and API session object

can probably go into its own package `@cennznet/api-react`

Aim is to reduce boiler plate for dapp developers
https://github.com/cennznet/cennzx-ui/blob/047a1e88d5841289696890f6f32e7aee62ad15aa/src/redux/reducers/extension.reducer.ts#L6-L55

```js
function App() {
    // connect api object
    // connect with extension
    // get extension addresses
    const {  cennznet, accounts } = useCENNZ('My App', { network: 'rata' });
    return <>
}
```

suggested by @miketamis

eupston:
closes #324

Will ideally be using this to refactor NFT Demo


packages/api/src/util/useCennznet.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (extensions.length === 0) {
+    return { api: api, accounts: null };
+  }
+  const polkadotExtension = extensions.find((ext) => ext.name === 'polkadot-js');
+  const metadata = polkadotExtension.metadata;
+  const checkIfMetaUpdated = localStorage.getItem(`EXTENSION_META_UPDATED`);
+  if (!checkIfMetaUpdated) {
+    const metadataDef = await extractMeta(api);
+    await metadata.provide(metadataDef);
+    localStorage.setItem(`EXTENSION_META_UPDATED`, 'true');
```
holygits, line 39: we should allow the extension to work with different networks at different times.

similar to what we did in cennzx something like 
`cennznet-ext-meta-{api.runtimeVersion.specName}-{api.runtimeVersion.specVersion}`
should be the key.

holygits (commented):
nice!

eupston:
Have tested in NFT Demo and works well.

Currently if the extension isn't installed `accounts` returns null, and if it is installed but no accounts are present than it returns an empty array. We may want to update the user docs at some point to highlight this detail, so devs can handle all situations appropriately on frontend.

packages/api/src/util/useCennznet.ts
eupston: sweet yeah good call, updated now. 
Decided to go with network name as key

packages/api/src/hooks/UseCennznet.ts
```diff
@@ -0,0 +1,78 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { web3Accounts, web3Enable } from '@polkadot/extension-dapp';
+import { InjectedAccountWithMeta, MetadataDef } from '@polkadot/extension-inject/types';
+import { getSpecTypes } from '@polkadot/types-known';
+import { defaults as addressDefaults } from '@polkadot/util-crypto/address/defaults';
+
+import { ApiOptions } from '../types';
+import { Api, Api as ApiPromise } from '../Api';
```
KarishmaBothara, line 21: will this only provide the promise version of API and not rx version?

packages/api/src/util/cennznetExtensions.ts
```diff
@@ -0,0 +1,53 @@
+import { InterfaceTypes } from '@polkadot/types/types';
```
KarishmaBothara, line 1: I think we can reuse this type from 
https://github.com/cennznet/api.js/blob/master/packages/types/src/interfaces/extrinsic/signedExtensions/cennznet.ts

packages/api/src/util/useCennznet.ts
KarishmaBothara: I think better to go with specVersion as for the same network, change in specVersion will introduce new metadata..

eupston: sweet updated now

packages/api/src/hooks/UseCennznet.ts
eupston: added option to return rx api as well

packages/api/src/util/cennznetExtensions.ts
eupston: o cool yup changed to use this type instead

packages/api/src/hooks/UseCennznet.ts
holygits: consider using another function instead of bool
e.g.
```ts
UseCennznet
UseCennznetRx
```

holygits (approved):
lgtm!
my only reservation is we should remove the metadata management stuff from here once extension PR
https://github.com/cennznet/extension/pull/9 lands and is published

packages/api/src/hooks/UseCennznet.ts
eupston: hmm yeah, probably more inline with our current API

eupston:
> lgtm!
> my only reservation is we should remove the metadata management stuff from here once extension PR
> [cennznet/extension#9](https://github.com/cennznet/extension/pull/9) lands and is published

yeah do you reckon we should wait until that PR is merged? or do it after the fact?

holygits:
I think we can merge this but wouldn't put it into an official release until metadata management is moved to extension, no harm in doing a pre-release with this change for testing purposes

packages/api/src/hooks/UseCennznet.ts
eupston: cool updated now

packages/api/src/hooks/UseCennznet.ts
```diff
@@ -0,0 +1,128 @@
...
+
+async function extractMeta(api) {
+  const systemChain = await api.rpc.system.chain();
+  const specTypes = getSpecTypes(
+    api.registry,
+    systemChain,
+    api.runtimeVersion.specName,
+    api.runtimeVersion.specVersion
+  );
+  if (specTypes.ExtrinsicSignatureV4) {
```
KarishmaBothara, line 88: can remove this delete specific class template with 
```

// remove all the classes from the spectypes
// the metadata that gets updated on the cennznet extension requires types of type (Record<string, string|object>)
                        const filteredSpecTypes = Object.keys(specTypes)
                            .filter(key => {
                                if (typeof specTypes[key] === 'function') {
                                    return false;
                                }
                                return true;
                            })
                            .reduce((obj, key) => {
                                obj[key] = specTypes[key];
                                return obj;
                            }, {});
```

KarishmaBothara: which would mean, in future even if we add new classes we won't need to add a delete that particular class line here..
The code will still work...

eupston: oh yup cool just updated

KarishmaBothara (approved):
lgtm

PR #343, opened 2021-07-06
Connects/cennznet 1.5.1 nikau

KarishmaBothara:
Closes https://github.com/cennznet/api.js/issues/342

API to connect to current develop/ image - cennznet/cennznet:1.5.1-nikau

Need to upgrade to @polkadot/api version which has those types that are introduced in this image.


packages/util/src/index.ts
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export * from '@polkadot/util';
-export * from '@polkadot/util-crypto';
+// export * from '@polkadot/util';
```
holygits, line 15: any action needed here?

packages/api/test/e2e/api.create.e2e.ts
```diff
@@ -27,6 +28,7 @@ describe('e2e api create', () => {
 
   beforeAll(async () => {
     await cryptoWaitReady();
+    await waitReady();
```
holygits, line 31: do we need both?
do users need to add this in their code?

holygits (commented):
thanks for this!

packages/util/src/index.ts
KarishmaBothara: I will re-add this export... my build used to fail when I had this...

KarishmaBothara: added it back and updated typescript version

packages/api/test/e2e/api.create.e2e.ts
KarishmaBothara: According to polkadot docs, it just needs one of them... however keyring was still not ready... I will play with this a bit and modify..

KarishmaBothara: what I observed is... 
for all our other test we create api instance first via 
`````
api = await initApiPromise();
`````
and then create the accounts with keyring
````
alice = keyring.addFromUri('//Alice');
````
Here api creation internally waits for waitReady from wasm-crypto and hence we don't need to add waitReady for others...

However in api creation test... we create api instance later and hence this was causing issue...

Have created the accounts after api instance is created in a test..

packages/util/src/index.ts
KarishmaBothara: have removed 
````
export * from '@polkadot/util-crypto';
````

The issue here is both of these packages internally has already exported a member name 'packageInfo'... which creates ambiguity... I couldn't find a way where I can export everything except one member... either export the list of member from every package over * or may be need a script to filter to filter packageInfo... will look at this

KarishmaBothara: https://github.com/Microsoft/TypeScript/pull/5148/files
This is introduced by typescript to avoid ambiguity when member of same name exist in the exported package...
exported individual package from @polkadot/util-crypto..

holygits (approved):

PR #351, opened 2021-08-01
support governance module branch

KarishmaBothara:
Can connect to governance branch (https://github.com/cennznet/cennznet/tree/feat/governance)
Added test for governance module


holygits (approved):

packages/api/test/e2e/api.v36-compat.e2e.ts
```diff
@@ -114,7 +114,7 @@ describe('runtime v36 compatibility', () => {
             done();
         });
 
-        it('Emits events when storage changes', async done => {
+        it.skip('Emits events when storage changes', async done => {
```
KarishmaBothara, line 117: Need to fix this.. for older runtime version ga creation fails with Invalid Transaction: BadProof error.. 
have 
````
types: [
        {
          minmax: [0, 36],
          types: {
            DispatchClass: 'DispatchClassTo36',
            DispatchInfo: 'DispatchInfoTo36',
            // In our staking runtime module, we use StakingLedgerTo223 definition of polkadot's staking ledger
            StakingLedger: 'StakingLedgerTo223',
            ExtrinsicSignatureV4: CENNZnetExtrinsicSignatureV0,
            ExtrinsicPayloadV4: CENNZnetExtrinsicPayloadV0,
            AssetInfo: 'AssetInfoV40'
          },
        },
        {
          minmax: [37, 40],
          types: {
            ...shareType37Onwards,
            AssetInfo: 'AssetInfoV40'
          },
        },
````
where AssetInfoV40 is of type

```
export interface AssetInfo extends Struct {
  readonly symbol: Bytes;
  readonly decimalPlaces: u8;
  readonly existentialDeposit: u64;
}

/** @name AssetInfoV40 */
export interface AssetInfoV40 extends Struct {
  readonly symbol: Bytes;
  readonly decimalPlaces: u8;
}
```
need some investigation here

Comment (KarishmaBothara) [reply packages/api/test/e2e/api.v36-compat.e2e.ts]: fixed

File: packages/api/test/e2e/governance.e2e.ts [hunk 2]
```diff
@@ -0,0 +1,168 @@
...
+    expect(members.includes(councilMemberAlice.address)).toBeTruthy();
+    expect(members.includes(councilMemberCharlie.address)).toBeTruthy();
+    expect(members.includes(councilMemberDave.address)).toBeTruthy();
+    done();
+  });
+
+  it('creates a proposal to add new council member', async done => {
+    const newCouncilMember = '5FWEHQqYMN8YCg8yJxKHnon7Dtx4Psp2xnjvKfQqGC6kUwgv';
+    const call = api.tx.governance.addCouncilMember(newCouncilMember);
+    const proposalCall = api.registry.createType('Call', call).toHex();
```
holygits, line 86: out of interest, does this work too?
```suggestion
    const proposalCall = api.tx.governance.addCouncilMember(newCouncilMember).toHex();
```

holygits (approved):

packages/api/test/e2e/governance.e2e.ts
KarishmaBothara: no.. it didn't work.. generates a different hex

packages/api/test/e2e/governance.e2e.ts
```diff
@@ -0,0 +1,168 @@
...
+    const justificationUri = 'https://example.com/nft/metadata'
+    const enactmentDelay = 26; // execute after 26 blocks once sufficient votes criteria is met
+    await api.tx.governance.submitProposal(
```
eupston, line 89: might be nice at some point if we have a wrapper method which includes adding council member and creating the proposalCall, so people just need to put in council members pub key.

eupston (approved):
lgtm!

packages/api/test/e2e/governance.e2e.ts
KarishmaBothara: plan is initially the council members would be created using sudo, then sudo would be set to null.. meaning no sudo..
After this adding a new council member, will be done only via council member by submitProposal - proposal would be to add a new council member..

KarishmaBothara: in the UI it will look like
Screen Shot 2021-08-11 at 12 20 51 PM
so the call is not just about adding council member, it can be anything

PR #352, opened 2021-08-02
Update hook

KarishmaBothara:
Updated hook as per new extension release.
We won't need DApp or api ask to update metadata for new runtime version.
Extension will take care of it..


packages/api/src/hooks/UseCennznet.ts
```diff
@@ -13,10 +13,7 @@
 // limitations under the License.
 
 import { web3Accounts, web3Enable } from '@polkadot/extension-dapp';
```
holygits, line 15: should publish the @cennznet/ versions sometime soon

holygits (approved):
lgtm!

packages/api/src/hooks/UseCennznet.ts
KarishmaBothara: ya to npm registry..

PR #356, opened 2021-08-11
Releasing version 2.0.0-alpha.4

KarishmaBothara:


packages/api/package.json
```diff
@@ -11,7 +11,7 @@
   "engines": {
     "node": ">=14.0.0"
   },
-  "main": "src/index.ts",
+  "main": "index.js",
```
holygits, line 14: still need this change?

scripts/generateSlimMetadataWrapper.js
```diff
@@ -2,7 +2,7 @@
 // Copyright 2017-2019 @polkadot/types authors & contributors
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
-
+process.env.JEST_WORKER_ID = '123';
```
holygits, line 5: is this change helpful?

KarishmaBothara: yes.. without this, generation of slim meta data fails..

packages/api/package.json
KarishmaBothara: reverted

PR #365, opened 2021-09-17
Deposit claim helper

KarishmaBothara:


packages/api/test/e2e/ethBridge.e2e.ts
```diff
@@ -165,13 +195,6 @@ describe('Eth bridge test', () => {
       done();
     });
 
-    it('Queries generic asset balance for test token 1 for BridgeTests account', async done => {
```
KarishmaBothara, line 168: balance is verified at line https://github.com/cennznet/api.js/pull/365/files#diff-f1bb908fb5b8ebb0dc7b51bdde1abe493f8632c8981ca21836dbaff0a990e47eR66

packages/api/src/util/helper.ts
```diff
@@ -27,3 +31,67 @@ export function extractEthereumSignature(signatures: EthereumSignature[]): { r:
...
+  return new Promise((resolve, reject) => {
+    api.tx.erc20Peg
+      .depositClaim(depositTxHash, claim)
+      .signAndSend(signer, { nonce }, ({ status, events }) => {
+        let eventClaimId;
+        if (status.isInBlock) {
+          for (const {
+            event: { method, section, data },
+          } of events) {
+            if (section === 'erc20Peg' && method === 'Erc20Claim') {
```
holygits, line 52: check event is really from the sender

packages/api/src/util/helper.ts
```diff
@@ -27,3 +31,67 @@ export function extractEthereumSignature(signatures: EthereumSignature[]): { r:
...
 }
+
+// Helper function to deposit claim that will return when claim is successfully deposited
+export async function awaitDepositClaim(
+  api: Api,
+  depositTxHash: H256 | string | Uint8Array,
+  claim: { tokenAddress: string; amount: string; beneficiary: string },
+  signer: KeyringPair
+): Promise<ClaimDeposited> {
+  const nonce = await api.rpc.system.accountNextIndex(signer.address);
```
holygits, line 42: do we need nonce separate?

packages/api/src/util/helper.ts
```diff
@@ -27,3 +31,67 @@ export function extractEthereumSignature(signatures: EthereumSignature[]): { r:
...
+              eventClaimId = data[0];
+              break;
+            } else if (section === 'system' && method === 'ExtrinsicFailed') {
+              reject('Claim already notarized');
+            }
+          }
+        }
+        if (eventClaimId) {
+          api.query.system
+            .events((events) => {
```
holygits, line 62: why query events again?

KarishmaBothara: to search for Erc20Deposit/Erc20DepositFail event... and then resolve the promise cause the deposit has already completed/failed

KarishmaBothara: When we check events for the first time.. those events are only generated via sign and send.. It will only get events related to depositClaim extrinsic... the second time, events are generic from everywhere..

PR #373, opened 2021-10-20
Added Mass Drop types and e2e test

JasonTulp:
Mass Drop types have been added with an end to end test to create and enter a mass drop


KarishmaBothara:
@JasonTulp can we close this PR - as per this commit https://github.com/cennznet/cennznet/commit/928509f4b8fc2002fd3474a97465902f4a32ed97?

PR #379, opened 2021-11-11
Fix derive type

KarishmaBothara:
Fixes ts error when using derived query..
Updated Ethy (event proof) call

### DIFF
--- a/packages/api/src/derives/ethBridge/eventProof.ts
+++ b/packages/api/src/derives/ethBridge/eventProof.ts
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import { extractEthereumSignature } from '@cennznet/api/util/helper';
+import { hexToString } from '@polkadot/util';
 import { Observable, combineLatest } from '@polkadot/x-rxjs';
 import type { ApiInterfaceRx } from '@polkadot/api/types';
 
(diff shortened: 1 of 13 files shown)


packages/api/src/derives/ethBridge/eventProof.ts
```diff
@@ -45,6 +46,8 @@ export function eventProof(instanceId: string, api: ApiInterfaceRx): () => Obser
             return {
               eventId: eventProof.eventId.toString(),
               validatorSetId: eventProof.validatorSetId.toString(),
+              blockHash: eventProof.blockHash.toString(),
+              tag: hexToString(eventProof.tag.toString()),
```
holygits, line 50: is this ok if tag is None?

holygits (approved):

packages/api/src/derives/ethBridge/eventProof.ts
KarishmaBothara: will check

KarishmaBothara: ya.. it is fine.. tried a small snippet with

```
const a = api.registry.createType('Option<VersionedEventProof>', null);
console.log('a:::', a.toString());
```

It returns - console.log
    a:::

Issue #403, opened 2021-12-20
case sensitive type definition issue


KarishmaBothara:
`the TS types are off, the typings say seriesMetadataUri but the actual fn is seriesMetadataURI`

KarishmaBothara:
https://github.com/polkadot-js/api/blob/6dd019343dcda48a4259a494904059fa0784f787/packages/typegen/src/generate/query.ts#L111
Polkadot script which creates the type definition... gets the name from the metadata and converts it to camelcase

https://github.com/polkadot-js/common/blob/e97410df5244a2de4bd3d0399f8e0aba0135fe6a/packages/util/src/string/camelCase.ts

PR #404, opened 2021-12-20
Fix definition with camel case storage for example seriesMetadataURI

KarishmaBothara:
Temporary fixes https://github.com/cennznet/api.js/issues/403
we will revert this change once https://github.com/cennznet/cennznet/issues/542 is solved

### DIFF
--- a/packages/types/src/interfaces/types.ts
+++ b/packages/types/src/interfaces/types.ts
@@ -1,11 +1,6 @@
 // Auto-generated via `yarn polkadot-types-from-defs`, do not edit
 /* eslint-disable */
 
-import {Api as ApiPromise, ApiRx} from "@cennznet/api";
-import {ApiTypes} from "@cennznet/api/types";
-import {AssetId} from "@cennznet/types/types";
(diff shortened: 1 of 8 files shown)


packages/types/src/interfaces/types.ts
```diff
@@ -1,11 +1,6 @@
 // Auto-generated via `yarn polkadot-types-from-defs`, do not edit
 /* eslint-disable */
 
-import {Api as ApiPromise, ApiRx} from "@cennznet/api";
-import {ApiTypes} from "@cennznet/api/types";
-import {AssetId} from "@cennznet/types/types";
-import {SubmittableExtrinsic as SubmittableExtrinsicBase} from "@polkadot/api/submittable/types";
-
```
KarishmaBothara, line 8: Moved the changes from here to index.js (one level up)
as every time we build/generate typedef, the types.ts will be overwritten

holygits (approved):

eupston (approved):
lgtm

PR #431, opened 2022-03-03
audit toHuman usage in the api

KarishmaBothara:
This PR is to audit usage of toHuman calls, as it can sometimes cause issue.

### DIFF
--- a/packages/api/src/derives/nft/openCollectionListings.ts
+++ b/packages/api/src/derives/nft/openCollectionListings.ts
@@ -124,9 +124,9 @@ export function openCollectionListingsV2(instanceId: string, api: ApiInterfaceRx
                       .map((tokenId) => {
                         return { listingId: listingId.toString(), tokenId };
                       })
-                      .flat();
+                      .reduce((acc, curr) => acc.concat(curr), []);
                   })
-                  .flat();
+                  .reduce((acc, curr) => acc.concat(curr), []);
                 const listingDetails = [];
                 // multi will make the query faster
                 return combineLatest([
--- a/packages/api/src/derives/nft/openCollectionListings.ts
+++ b/packages/api/src/derives/nft/openCollectionListings.ts
@@ -39,7 +39,7 @@ export function openCollectionListings(instanceId: string, api: ApiInterfaceRx) 
           const listingIDs = storageKeys.map((storageKey) => {
             return storageKey.args.map((k) => k.toString())[1];
           });
-          return api.query.nft.listings.multi(listingIDs.map((listingID) => listingID)).pipe(
+          return api.query.nft.listings.multi(listingIDs).pipe(
             switchMap(
               (listingsCodec: Option<Listing>[]): Observable<DeriveTokenInfo[]> => {
                 const tokenIds = listingsCodec.map((listingCodec) => {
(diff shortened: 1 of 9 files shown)


packages/api/src/derives/nft/openCollectionListings.ts
```diff
@@ -89,3 +90,74 @@ export function openCollectionListings(instanceId: string, api: ApiInterfaceRx)
...
+                      .reduce((acc, curr) => acc.concat(curr), []);
+                  })
+                  .reduce((acc, curr) => acc.concat(curr), []);
+                const listingDetails = [];
+                // multi will make the query faster
+                return combineLatest([
+                  api.query.nft.tokenOwner.multi(
+                    queryArgs.map(({ tokenId }) => [[tokenId[0], tokenId[1]], tokenId[2]])
+                  ),
+                  api.query.nft.seriesAttributes.multi(queryArgs.map(({ tokenId }) => [tokenId[0], tokenId[1]])),
```
holygits, line 136: since we're deprecating onchain attributes readonly, can skip this query

packages/api/test/e2e/api.create.e2e.ts
```diff
@@ -45,6 +45,13 @@ describe('e2e api create', () => {
     done();
   });
 
+  it('Find tokens listing on nkau', async done => {
```
holygits, line 48: need a better test since this will become meaningless once the listings close

packages/api/src/derives/nft/openCollectionListings.ts
```diff
@@ -89,3 +90,72 @@ export function openCollectionListings(instanceId: string, api: ApiInterfaceRx)
...
 }
+
+/**
+ * Gets all tokens in a collection that have an open listing
+ *
+ * @param collectionId  The collection Id value
+ *
+ * @returns List of tokens that have open listings in given collection in the format {listingId, tokens: vec<DeriveTokenInfo>}
+ */
+export function openCollectionListingsV2(instanceId: string, api: ApiInterfaceRx) {
```
ken-centrality, line 101: so this new version won't bring back the `attributes` anymore? in that case for older runtime version, they have to do another query to return the `attributes`?

packages/api/test/e2e/nft.e2e.ts
```diff
@@ -482,6 +481,13 @@ describe('NFTs', () => {
       })
   });
 
+
+  it('Find tokens listing on local with V2', async done => {
+    const allTokens = await api.derive.nft.openCollectionListingsV2('15');
+    expect(allTokens.length).toBe(2);
```
ken-centrality, line 487: can we do more `expect` here for the details of each tokens?

ken-centrality (changes requested):

packages/api/src/derives/nft/openCollectionListings.ts
KarishmaBothara: since we're deprecating onchain attributes, can skip this query.. for older chain, they would need to query at that blockHash.. when the runtime is upgraded the call to get attributes will fail, api.query.nft.seriesAttribute as there s no more seriesAttribute in the storage

packages/api/test/e2e/nft.e2e.ts
KarishmaBothara: yes.. will add more

packages/api/src/derives/nft/openCollectionListings.ts
ken-centrality: Yup, that's fine, you might want to note that down for the upgrade guide

ken-centrality (approved):

Issue #400, opened 2021-11-25
Missing type definitions for NFT rpc methods


holygits:

KarishmaBothara:
I think this is solved with  https://www.npmjs.com/package/@cennznet/api/v/2.1.0-alpha.1 version..
I will add test for these soon and close this issue.

Issue #375, opened 2021-11-02
Reenable the test for Nikau/Rata once they are back


KarishmaBothara:

holygits:
they're back!

PR #458, opened 2022-04-05
rpc type definition + enable test net

KarishmaBothara:
Desc - Fixes rpc definition and enable test for rata..
closes - https://github.com/cennznet/api.js/issues/375 and https://github.com/cennznet/api.js/issues/400


eupston (approved):
lgtm

PR #471, opened 2022-05-10
Feat/encode eth signature

KarishmaBothara:
Compatible with https://github.com/cennznet/cennznet/pull/639


KarishmaBothara:
the test will pass, only when we update with new docker image from node with the changes

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
```diff
@@ -95,8 +96,9 @@ export default class CENNZnetExtrinsic extends GenericExtrinsic{
     const cennznetAddress = cvmToAddress(ethAddress);
     const nonce = await api.rpc.system.accountNextIndex(cennznetAddress);
     const payload = this.registry.createType('EthWalletCall', { call: this, nonce }).toHex();
+    const encodedPayload = `data:application-octet;base64,${base64Encode(payload)}`;
```
ken-futureverse: should be `data:application/octet-stream`

ken-futureverse (changes requested):

packages/types/src/interfaces/extrinsic/v1/Extrinsic.ts
KarishmaBothara: ya was waiting for image from node changes.. updated prefix and signature with change in prefix

packages/api/src/util/helper.ts
```diff
@@ -48,15 +50,25 @@ export async function awaitDepositClaim(
...
+            } else if (
+              (section === 'system' && method === 'ExtrinsicFailed') ||
+              section === 'erc20Peg' ||
+              method === 'DelayedErc20DepositFailed'
+            ) {
               reject('Claim already notarized');
             }
           }
         }
+        console.log('event claim id:', eventClaimId);
```
ken-futureverse: remove the `console.log`?

ken-futureverse (approved):

